PR #155, opened 2026-09-12
feat(pr-review-loop): local PR review loop, driven by git's own record of what was pushed

yqrashawn:
A Claude Code plugin that reviews every agent push to a PR, locally, and wakes
the pushing session with the findings. Replaces a GitHub Action that billed
Actions minutes on a private repo and could not see the working tree, the
history, or a running dev stack.

## How it decides

One question, for both `git push` and `gh pr create`:

> is there a push, **made by an agent** and **landed**, whose sha is the head
> of an open PR with **no ledger row**?

- **made by an agent** — a `pre-push` hook records `CLAUDE_CODE_SESSION_ID`;
  a human's terminal push records `-`. That is R2, decided by what made the
  push rather than by how the command was spelled.
- **landed** — git's own remote-tracking reflog, written only on success.

The reviewer runs as a separate `claude -p` in a throwaway worktree pinned to
the reviewed sha, so it cannot see the author's conversation and the tree
cannot move under it. State lives in `<git-common-dir>/pr-review*`: an
append-only ledger, one lock per PR, the context diff, the findings text.

## The approach this replaces

The first design read the shell command to find which directory the push ran
in. That produced five separate defects over two days — a worktree the session
cwd did not name, `gh pr create` skipped for its heredoc, rtk silent on
heredocs, `$(...)` before the push, a two-level variable chain — and, measured
against a corpus of 260 real push commands extracted from 1888 transcripts,
still missed 25 of them.

It is withdrawn entirely. Git already records every push exactly, however the
command was written. Constraints C42–C47 in the spec stand as the record of
why that approach was abandoned rather than as description of the code.

## Evidence

Three days of real use across four clones — `claude-code-http-proxy`,
`neobimo`, `perpdex-chain-scan`, `.nixpkgs`:

- **132 completed reviews**, 17 of them `NOT MERGEABLE`
- **perpdex: 14/14** PRs had their final head reviewed; cchp 28/33; neobimo 8/10
  (the two misses are the pass cap firing, by design)
- 272 fingerprints across cchp's 50 rows, median 5.5 findings per pass, zero
  empty rows
- the killed-review retry fired 7 times and recovered a review that had been
  killed mid-run

## Known open items

- **Follow-up findings do not converge.** Each pass re-reviews the whole
  accumulated diff rather than the increment, and 62% of findings on passes
  2–10 land on ground already reviewed and passed over. Fixing findings grows
  the diff, which grows the pool. Two PRs hit the 10-pass cap. The cap is
  currently the only damping term.
- Versions 0.23/0.24 raise the bar for non-correctness findings and are
  **untested** — no session has restarted onto them yet.
- One reviewer death remains unexplained (cchp #410: 20 minutes, zero bytes of
  stderr). Stderr is now streamed to disk so the next one is diagnosable.

## Not part of this work

Three commits are unrelated drive-bys kept only because they are already in the
branch: `1ccf6b856` (lemmalog MCP server) and `c825a2ef1` (codebase-memory-mcp
runtime dir), both touching `.doom.d/mcp.el`, and `208922923` (remove the
`permission_prompt` MCP tool), whose fallout is ~250 deleted lines across four
`.doom.d` files — the dead tests and the `b2.org` E2E section that documented
the removed feature. None of the removed elisp symbols is referenced anywhere
under `.doom.d/`, and the removal is pinned by an assertion that the
`--permission-prompt-tool` flag never comes back.

## Verification

270 tests, 760 assertions, 0 failures (`bb test` in `claude-code-plugins/`),
green with `PR_REVIEW_LOOP_REVIEWER` both set and unset.
Every behavioural fix in this branch carries a negative control — the test was
shown to fail with the fix reverted — because several early tests passed with
and without their fix.

## Summary by CodeRabbit

* **New Features**
  * Added the `pr-review-loop` Claude Code plugin for automated reviews after pushes.
  * Reviews run against the pushed commit, post findings to the PR, and notify the active session.
  * Added on-demand reviews, safeguards, pass tracking, retries, and linked-worktree support.

* **Configuration**
  * Added marketplace metadata, setup automation, MCP server configuration, and runtime-directory support.

* **Documentation**
  * Added plugin usage, troubleshooting, review guidance, and design documentation.

* **Changes**
  * Removed the Claude Code permission-prompt workflow.


yqrashawn:
## pr-review-loop — pass 1

`2d4a51df1905` — **NOT MERGEABLE**

Findings below: **verified by RUNNING** — #1, #2, #3, #5, #7, #9 (partly), #11. Everything else is a reading of the source. All probes ran in this throwaway worktree against temp directories; nothing touched the user's real `pushes.log`, ledger, or the PR.

One environment note so it is not mistaken for a finding: `bb test` failed 12 assertions with 2 errors on my first run *because my own environment has `PR_REVIEW_LOOP_REVIEWER=1`*. With it unset the suite is **239 tests, 681 assertions, 0 failures, 0 errors**. That is not a defect in the diff — but it is finding #7.

VERDICT: NOT MERGEABLE — a re-review that quotes the prior pass's PR comment has that quote parsed as its own verdict, recording MERGEABLE over a NOT MERGEABLE review with a blocking finding

  [correctness/blocking]  2
  [correctness/followup]  2
  [coverage]              3
  [docs-accuracy]         5
  [style]                 1

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/reviewer.clj:245 — `parse-verdict` takes the LAST column-0 `VERDICT:` and `parse-counts` (:331) reads the block after it, so a quoted prior verdict beats the real one. Traced and RUN: a pass-2 reply whose own verdict is `NOT MERGEABLE` with `[correctness/blocking] 1`, which then quotes pass 1's comment in a ``` fence (a fence does not indent), parses as `MERGEABLE`/`0 blocking` — `reconcile` then confirms it, the ledger row records MERGEABLE, and `findings-message` (trigger.clj:355) tells agent A "you may merge this PR now". This is created by this branch: HEAD (2d4a51df1) posts every pass as a PR comment, prompt.clj:69 instructs the reviewer to read them, and review_core.md:140 tells a re-review to verify closure against the previous pass. The prompt's only guard (review_core.md:31-34) covers echoing *the prompt* and restating the format *before* reviewing — neither covers quoting a comment after the verdict. Reviewers recapping prior passes is already measured (reviewer.clj:320-325).
2. [correctness/blocking] claude-code-plugins/hooks/pr_review/attempts.clj:137 — `prune!` rewrites `pushes.log` with `spit` (truncate-in-place, no atomic rename) under a flock that the POSIX `pre-push` hook's bare `>>` (pr-push-hook.sh:45,57) never takes. Two consequences: a record appended during the read→write window is destroyed — RUN, 165 of 400 concurrent appends lost, each a push that can then never match a reflog entry and is silently never reviewed — and a kill between truncate and write zeroes the whole file, losing every clone's push provenance. This is the exact shape `ledger/append-pass!` fixed with temp+`ATOMIC_MOVE` (ledger.clj:125-135); the sweep finds two more unfixed siblings, both writing files the user owns: hookinstall.clj:130 (`spit` of `pre-push` itself, chmod only afterwards — a truncated executable hook aborts the user's push, violating that file's own stated never-fail-a-push invariant) and hookinstall.clj:111 (`info/exclude`, read-modify-write with no lock while `-main` loops over every known clone).
3. [coverage] claude-code-plugins/test/pr_review/reviewer_test.clj:507 — `the-token-never-appears-in-what-the-loop-records` certifies a credential-non-leak safety property it does not check: its stub child is `echo out; echo err >&2`, which has no reason to print the token, so the test cannot fail. There is no redaction anywhere in the tree (grepped). RUN against the real `default-spawn` and `parse-output` with a child that does print `$CLAUDE_CODE_OAUTH_TOKEN`: the value lands in `:out`, `:err`, the streamed stderr file, and `(:body parsed)` — and `(:body parsed)` is what trigger.clj:325/504 sends to `gh pr comment` and trigger.clj:498 spits to the findings file. The test names those artifacts and checks none of them.
4. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:186 — the long-lived OAuth token is placed in the environment of a `claude -p` that has Bash unrestricted (reviewer.clj:118 denies no `gh`, no shell), and the loop pipes that process's entire stdout to a GitHub comment with no filtering. The plumbing is traced and RUN (see #3); the trigger — the reviewer choosing to emit the value, e.g. running `env` while debugging — is hypothesised, which is why this is followup rather than blocking.
5. [correctness/followup] claude-code-plugins/hooks/pr_review/trigger.clj:264 — `decide-outside-a-reviewer` builds the decision without `:pr-url` or `:pushed-by`, so on every hook-triggered review the prompt's "What the author says this change does" section is silently absent (prompt.clj:68 gates on it) and the wake never names the pushing session (trigger.clj:343). RUN: `decide` with a stubbed `open-pr` returning `:url` yields keys `(:action :base-ref :branch :candidates :draft? :git-dir :pass :pr :prior-fingerprints :repo-root :sha :trigger)` — no `:pr-url`, no `:pushed-by`. `gh/open-pr` requests `url` for exactly this purpose (gh.clj:99-102) and `attempts/pusher` (attempts.clj:104) was written for exactly the other, but only manual.clj:75 wires the first and nothing calls the second. The `:pushed-by` gap matters more than it looks: `abandoned-candidates` is deliberately neither session- nor verb-scoped, so a wake routinely lands in a session that did not push.
6. [coverage] claude-code-plugins/test/pr_review/prompt_test.clj:85 — `the-prompt-points-the-reviewer-at-the-pr` passes `:pr-url` straight to `prompt/build`, and trigger_test.clj:540 passes `:pushed-by` straight to `findings-message`; neither can fail for the reason it claims, because the automatic path never supplies either key. Sibling of #5 — the tests are why the gap survived.
7. [coverage] claude-code-plugins/test/pr_review/trigger_test.clj:891 (siblings :971, :1016) — these three end-to-end tests do not stub `:reviewer-env-fn`, so `decide` short-circuits to `:silent` whenever `PR_REVIEW_LOOP_REVIEWER` is set in the ambient environment. RUN both ways: 12 failures / 2 errors with it set, 0 with it unset. review_core.md:88 permits the reviewer to run the suite, so the loop's own reviewer on this repo gets a red suite it is told not to attribute to the diff.
8. [docs-accuracy] claude-code-plugins/README.md:42 — "the reviewer … has `Bash(gh pr:*)` denied" is false, and so are its three siblings: README.md:94 ("`gh pr` … denied as command shapes"), skills/pr-review-loop/SKILL.md:81 ("It cannot alter the PR: `git push`, `git commit` and `gh pr` are refused"), and hooks/pr_review/trigger.clj:500 ("`Bash(gh pr:*)` stays denied to it"). `denied-tools` (reviewer.clj:103-118) contains no entry matching `gh`, its own docstring says so in capitals, and reviewer_test.clj:75 asserts it. The reviewer can merge, close or comment on the PR it is reviewing — SKILL.md is the one agent A reads.
9. [docs-accuracy] claude-code-plugins/hooks/review_core.md:91 — "**Do not read CI.** You have no network." The reviewer has `gh`, and line 69 of the same file tells it to run `gh pr view … --comments`; I used that successfully in this review. Sibling at reviewer.clj:78, which says "The prompt asks the reviewer not to [run the test suite]" while review_core.md:88 allows it.
10. [docs-accuracy] claude-code-plugins/hooks/pr-push-hook.sh:4 — "Records WHICH CLONE was pushed, and deliberately nothing else", and lines 6-12 argue that recording refs or shas here "would mean recording pushes that were then rejected". Lines 52-60 record one ref and sha per push. The later comment at 47-51 gives the real reasoning; the header was not updated with it.
11. [docs-accuracy] .doom.d/modules/llm/gptel-claude-code/gptel-claude-code-mcp.el:1 — the PR body's "Not part of this work" names two unrelated drive-by commits (`1ccf6b856`, `c825a2ef1`). There is a third: `208922923 refactor(gptel-claude-code): remove the permission_prompt MCP tool`, 254 deletions across three `.doom.d/modules/llm/gptel-claude-code/` files, unmentioned anywhere in the body.
12. [docs-accuracy] claude-code-plugins/bb.edn:21 — the PR body's "Verification: 200 tests, 532 assertions, 0 failures (`bb test`)" is stale; at HEAD it is 239 tests, 681 assertions. RUN. The "0 failures" half holds.
13. [style] claude-code-plugins/hooks/pr_review/hookinstall.clj:102 — the second destructuring rebinds `exit` and `out` in the same `let` that already bound them at :99, then uses neither (it reads `(:exit ce)`/`(:out ce)` instead). The shadowed names are dead, but a reader has to prove that before trusting `top`.

yqrashawn:
## pr-review-loop — pass 2

`e0718a85c41e` — **NOT MERGEABLE**

Read the diff, the three fix commits, and the surrounding source; ran two targeted reproductions and the test suite twice.

**Evidence basis:** findings #1, #2 and #3 were verified by **RUNNING** (a scripted end-to-end call into `trigger/review!` against a temp clone, and the suite with and without `PR_REVIEW_LOOP_REVIEWER`). Findings #4–#9 are a reading of the source, with the false-doc claims checked against the code they describe. Everything ran in this throwaway worktree against temp directories; nothing touched the real ledger, `pushes.log`, or the PR.

**Closure of pass 1's blocking and coverage findings:**
- #1 (quoted verdict parsed as this one) — **closed**. `reviewer.clj:265` `unfenced` blanks fenced regions before `strip-emphasis` (`normalized-lines`, :289), covering verdict, counts and fingerprints; three tests, including the ordering one.
- #2 (`prune!` truncating `pushes.log`) — **closed**. `prune!` is deleted; `tail-lines` (attempts.clj:83) bounds the reader instead, so the log has no truncating writer.
- #3 (token test could not fail) — **closed**. `reviewer_test.clj:516` now has the stub child print the credential and asserts `[redacted]` rather than absence; `redact` runs in `run!` on `:out`, `:err` and the stderr file.
- #7 (ambient `PR_REVIEW_LOOP_REVIEWER`) — **not closed**, at the same three sites. Re-raised below as #3.

VERDICT: NOT MERGEABLE — `run-review!` never binds `pr`, so the new PR-scoped checkout path is built from `clojure.core/pr`; two PRs at one sha still share a directory and `prune-stale!` can no longer match any real checkout

  [correctness/blocking]  1
  [correctness/followup]  3
  [coverage]              2
  [docs-accuracy]         3
  [style]                 none

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/trigger.clj:529 — `run-review!` destructures `{:keys [git-dir sha] :as d}` (:527) and then passes a bare `pr` to `with-checkout`. `pr` is not bound there, so it resolves to `clojure.core/pr` and the path becomes `pr-review-sci.impl.io$pr@78cc4ec-<sha12>`. RUN, end to end through `#'trigger/review!` on a temp clone: PR 401 and PR 402 at one sha both got `.../pr-review-sci.impl.io$pr@78cc4ec-ffc7e16b779b` — `SAME DIRECTORY FOR TWO PRs AT ONE SHA? true`. That is precisely the defect `e0718a85c` says it fixed; since the lock became per PR those two reviews run concurrently and `add!` (checkout.clj:114-116) unconditionally `worktree remove --force` + `delete-tree`s the path on entry, so the second review deletes the first's tree mid-run and the first's `remove!` deletes the second's. Second consequence, also RUN: the name does not match `review-dir-re` (checkout.clj:55) — `re-matches` → `false` — so `review-worktrees` never lists a real checkout and `prune-stale!` can never reclaim one; a killed review's surviving directory and its `.git/worktrees/<name>` entry leak permanently in the user's clone. The value is stable across babashka processes (three runs, all `sci.impl.io$pr@78cc4ec`), so this is not intermittent. `manual.clj:88` reaches the same code through `trigger/respond`. Siblings, same root: checkout.clj:95 ("The PR is in the path, not just the sha") and the commit subject "split checkouts by PR" are both false as wired. `clj-kondo` is clean on this file — a core-name collision is a legal reference, so no linter will catch the class.
2. [coverage] claude-code-plugins/test/pr_review/trigger_test.clj:92 — `:with-checkout-fn (fn [_gd _pr _sha _parent _opts f] ...)` discards the PR argument, so no test observes what `run-review!` actually passes. Its counterpart, `checkout_test.clj:219` `two-prs-reviewing-the-same-sha-do-not-share-a-directory`, certifies that safety property by calling `co/add!` with literal `401`/`402` — the property it names does not hold on the only production path, and the suite is green either way. RUN: `bb test` → 258 tests, 725 assertions, 0 failures, while the repro above shows the collision. `checkout_test.clj:118` shows the author already draws the function/wiring distinction ("the wiring, not just the function"); this is the case where it was not applied.
3. [coverage] claude-code-plugins/test/pr_review/trigger_test.clj:902 (siblings :979, :1036) — re-raise of pass 1's #7, unchanged. These three `trigger/decide` calls pass raw opts maps without `:reviewer-env-fn`, so `decide` (trigger.clj:305) short-circuits to `:silent` whenever the variable is set in the ambient environment. The shared `opts` helper gained the stub at :53; these did not. RUN both ways on `pr-review.trigger-test`: unset → 62 tests / 198 assertions / 0 failures; set → **12 failures, 2 errors**, in exactly `a-retry-survives-a-clone-index-entry-days-old`, `a-real-abandoned-lock-is-retried-end-to-end` and `a-real-clone-with-a-real-reflog-decides-a-review`. `review_core.md:88` permits the reviewer to run the suite, so this loop's own reviewer on this repo gets a red suite it is told not to attribute to the diff.
4. [correctness/followup] claude-code-plugins/hooks/pr_review/trigger.clj:193 — `abandoned-candidates` builds `{:git-dir … :branch … :new-sha … :ts … :retry? true}` with no `:session`, so `decide-outside-a-reviewer`'s `session (assoc :pushed-by session)` (:268) is nil on every retry. That is the one path `e0718a85c`'s own message names as the reason `:pushed-by` matters ("`abandoned-candidates` is deliberately neither session- nor verb-scoped, so a wake routinely lands in a session that did not push"); the fix covers the reflog path only. TRACED. `attempts/pusher` (attempts.clj:138), written for exactly this lookup, still has zero call sites. Both new tests (trigger_test.clj:1051, :1074) exercise the reflog path, so neither can see the gap. Effect is bounded: the wake omits the ownership sentence on retries.
5. [correctness/followup] claude-code-plugins/hooks/pr_review/trigger.clj:183 — "the index is pruned at 24h, which is the real bound on how long a retry stays reachable" is no longer true: `f05248f69` deleted `attempts/prune!`, and `attempts/max-age-ms` (attempts.clj:39) is now dead — grep finds only its definition, and clj-kondo reports `Unused private var`. `abandoned-candidates` calls `clones-since log 0`, so on every Bash tool call it now walks lock files in *every clone ever pushed from that still exists*, forever, rather than the last 24 hours' worth. Cost only, not corruption — `actionable` still gates each candidate on an open PR at that sha with no ledger row — but the stated bound is gone and nothing replaced it.
6. [correctness/followup] claude-code-plugins/hooks/pr_review/hookinstall.clj:142 — `install-into!`'s `write!` calls `atomicfile/spit!`, which writes a fixed sibling `<path>.tmp` (atomicfile.clj:28), and unlike `exclude-locally!` (:114) it holds no lock. `-main` (:236) loops over every known clone from SessionStart, so two sessions starting at once in one clone both open `<hooksdir>/pre-push.tmp` with `O_TRUNC`; if one truncates between the other's write and its `ATOMIC_MOVE`, a partial `pre-push` is renamed into place. `pr-push-hook.sh` runs under `set -u` with the record block inside `{ … }` (:27-65), so a cut mid-brace is a shell syntax error and a non-zero pre-push aborts the user's push — the one thing that file's header forbids by name. Narrower than the `spit`-then-`chmod` window it replaced, so this is a residue of the fix rather than a regression. TRACED; not raced.
7. [docs-accuracy] claude-code-plugins/hooks/pr_review/trigger.clj:510 — "`Bash(gh pr:*)` stays denied to it" survives the sweep that corrected README.md:42, SKILL.md:81 and spec C55/C56b. No entry matching that shape exists in `denied-tools`; only the seven mutating subcommands plus `gh api` do (reviewer.clj:120-122). Sibling, opposite direction: reviewer.clj:79-80 still says "The prompt asks the reviewer not to [run the test suite]", while review_core.md:88 says "Running the test suite is allowed but rarely worth it."
8. [docs-accuracy] claude-code-plugins/bb.edn:21 — the PR body's "Verification: 200 tests, 532 assertions, 0 failures (`bb test`)" is stale at HEAD; RUN, it is 258 tests and 725 assertions, matching `e0718a85c`'s own message but not the body. Same artifact, pass 1's #11 unclosed: the body's "Not part of this work" still names only `1ccf6b856` and `c825a2ef1`, while `208922923` — the permission-prompt removal, whose fallout took 203 more deleted lines in this HEAD commit — is still unmentioned there.
9. [docs-accuracy] .doom.d/modules/llm/gptel-claude-code/b2.org:46 — `e0718a85c` claims "Nothing in .doom.d references any of the six symbols", which holds for the symbols (grep: zero hits). But b2.org's Test 5 still walks the reader through the removed feature: the `permission_prompt` MCP tool (:46), the `*gptel-permission*` popup (:51, :66) and `--permission-prompt-tool mcp__emacs__permission_prompt` in `*Messages*` (:72). `gptel-claude-code-test.el:826` already asserts that flag is never passed, so the manual scenario documents something the suite now forbids.

yqrashawn:
## pr-review-loop — pass 3

`3a46512c61b1` — **NOT MERGEABLE**

Read the diff, the three fix commits, and the surrounding source. Ran the suite twice and three targeted probes.

**Evidence basis:** findings #1, #2 and the closure of pass 2's #3 were verified by **RUNNING** — two `bb` probes against `pr-review.context/build!` and `pr-review.reviewer/parse-output`, plus `bb test` with and without `PR_REVIEW_LOOP_REVIEWER`. Findings #3–#8 are a reading of the source, with each doc claim checked against the code it describes. Everything ran in this throwaway worktree against temp directories; nothing touched the real ledger, `pushes.log`, or the PR.

**Closure of pass 2's blocking and coverage findings:**
- #1 (`run-review!` passed an unbound `pr`) — **closed**. `trigger.clj:538` now destructures `{:keys [git-dir pr sha]}`, and `trigger_test.clj:1152` asserts `401` reaches the checkout through the real `review!`.
- #2 (the `with-checkout-fn` stub discarded its PR argument) — **closed**. `trigger_test.clj:94` records it into `last-checkout-pr`.
- #3 (ambient `PR_REVIEW_LOOP_REVIEWER`) — **closed**. All three sites now stub `:reviewer-env-fn`. RUN both ways: unset → 261 tests / 733 assertions / 0 failures; set → identical.

The same root cause pass 2 found is still live one call frame away: `review-in!` passes `:pr` to `context/build!`, which ignores it, so the diff file is keyed by sha alone.

VERDICT: NOT MERGEABLE — two open PRs at one sha share `<git-dir>/pr-review-context/<sha>.diff`, so one review's `spit` replaces the diff the other's reviewer was pointed at, and its findings are posted to the wrong PR

  [correctness/blocking]  2
  [correctness/followup]  1
  [coverage]              2
  [docs-accuracy]         2
  [style]                 1

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/context.clj:58 — `diff-path` is `<context-dir>/<sha>.diff`, keyed by sha only, while `build!` (:48) destructures `{:keys [sha base-ref]}` and silently drops the `:pr` that `review-in!` passes it at trigger.clj:461. This is pass 2's #1 one frame down: the checkout and the lock were split by PR, the context file was not. RUN, two `build!` calls at one sha with different base refs: `same path? true`, and PR 401's reviewer reads `diff --git a/PR402.clj b/PR402.clj` while PR 401's own prompt still lists `PR401.clj` under "Changed files" (`changed-files` comes from the in-memory text, the reviewer's bytes come from the file). The findings are then posted to PR 401 by trigger.clj:526. Reachability is the author's own, not hypothesised: checkout.clj:95-98 states two open PRs can share a head sha, and since lock.clj:22-35 made the lock per PR "their reviews now run CONCURRENTLY" — that is exactly and only when this collides. Sibling, same root, same pass: `context/prune!` (context.clj:68, called at trigger.clj:510) keeps the 5 newest `.diff` files across the whole clone, so a finishing review can delete the file a concurrently running review's prompt names. Two of the three consequences — wrong-PR content, and `spit`'s truncate window on a file another process is reading — need no concurrency beyond that documented case.

2. [correctness/blocking] claude-code-plugins/hooks/pr_review/reviewer.clj:298 — `unfenced` closed the fenced spelling of pass 1's #1, not the class. `parse-verdict` still takes the LAST column-0 `VERDICT:` among unfenced lines, so a prior pass's verdict quoted plainly — no fence, no indent, under a "Closure of the previous pass" heading, which is the shape review_core.md:140 asks a re-review to produce — beats the reviewer's own. RUN against `parse-output` + `reconcile` on a reply whose own verdict is `NOT MERGEABLE` with `[correctness/blocking] 1`, followed by a verbatim quote of a prior `MERGEABLE` comment: `verdict: MERGEABLE`, `correctness/blocking 0`, `mergeable? true`. The ledger row records MERGEABLE and `findings-message` (trigger.clj:388) tells agent A "MERGEABLE means you may merge this PR now". Note the direction: reviewer.clj:387-394 documents a reviewer recapping a previous pass as *measured behaviour*, and the fix for it (counts read after the winning verdict) is what makes the quoted block's counts authoritative here. Sibling, bounded: `parse-fingerprints` (reviewer.clj:371) deliberately reads the whole reply, so the quoted pass's finding lines enter this row's `:fingerprints` and feed `ledger/suppressed-fingerprints` (ledger.clj:118) — a follow-up finding raised once and quoted once is suppressed as if raised twice.

3. [coverage] claude-code-plugins/test/pr_review/context_test.clj:47 — `diff-path-is-namespaced-by-sha` certifies the safety property "one file per SHA so a superseded reviewer's diff is never overwritten under it mid-read" and checks only that the string equals `<context-dir>/abc123.diff`. Sha-only naming is precisely what *does* get overwritten mid-read once two PRs share a sha (#1). The test also passes `:pr 370` into `build!`, which ignores it — the same stub-discards-the-argument shape as pass 2's #2, this time in the assertion's input rather than in a fake.

4. [coverage] claude-code-plugins/test/pr_review/reviewer_test.clj:554 — `a-quoted-prior-review-is-not-parsed-as-this-one` names the general property and exercises one spelling: the quote is inside a ``` fence (:564). The unfenced quote fails (#1 of the two RUN probes). `fences-are-stripped-before-emphasis-not-after` (:596) and `a-fenced-code-excerpt-inside-a-finding-keeps-the-finding` (:583) are both fenced too, so nothing in the suite can observe the open variant, and `counts-come-from-the-winning-verdicts-own-block` (:370) places its recap *before* the real verdict — the safe side of the same asymmetry.

5. [correctness/followup] claude-code-plugins/hooks/pr_review/atomicfile.clj:44 — the `finally` deletes the unique temp on the *exception* path only. The path this namespace exists for is the kill path (its own docstring: "a process killed in between destroys it"), and a SIGKILL between `spit tmp` and `Files/move` leaves `<path>.tmp.<nanos>.<rand>` behind with no writer that will ever reuse or clean it. The comment at :46-48 asserts the opposite ("After a successful move there is nothing left to delete", offered as the reason the `finally` is sufficient). Bounded, and only visible where the hooks directory is inside a working tree — a relative `core.hooksPath` such as perpdex-chain-scan's `.githooks` (hookinstall.clj:77-80) — because `exclude-locally!` (:93) excludes `pre-push` and nothing else, so each leftover shows as untracked in the user's `git status` forever. The fixed name it replaced was self-healing here; the unique name trades the narrow tear for unbounded litter. TRACED, not raced.

6. [docs-accuracy] claude-code-plugins/hooks/pr_review/prompt.clj:84 — "`git push` and `git commit` are refused; `gh` is not, so do not comment on, merge or close the PR" is false as of `reviewer.clj:121-123`, and it is the copy that reaches the reviewer: this exact sentence is in the prompt I was given, two sections below review_core.md's corrected "every subcommand that changes it … is refused", which it contradicts. Two siblings the same sweep missed: `skills/pr-review-loop/SKILL.md:83` says "it has no network" in the same sentence that grants it `gh pr view`, which is a network call — and `git`, `git fetch` included, is not denied either; `docs/superpowers/specs/2026-09-08-pr-review-loop.md:128` (R7) still says "`git push`/`git commit`/`gh pr` are refused", which C56b at :108 of the same file exists to correct.

7. [docs-accuracy] claude-code-plugins/hooks/pr_review/trigger.clj:647 — "Housekeeping, last of all: … nothing `prune!` does — or fails to do — can reach the review. It sits outside every branch `decide` can take … and it is the only thing that ever deletes these files." `-main` has no `prune!` call; `attempts/prune!` was deleted in `f05248f69` and the comment survived it. The only statement the comment now guards is `(System/exit exit)`.

8. [style] claude-code-plugins/hooks/pr_review/attempts.clj:138 — `pusher`'s docstring ends mid-sentence: "Either way someone has `since-ms` 0 deliberately." The HEAD commit replaced the clause that finished it ("to be told whose PR it is, and this file already knows") with the new paragraph and left the stump attached to the front of it.

yqrashawn:
## pr-review-loop — pass 4

`00640a275b26` — **NOT MERGEABLE**

## pr-review-loop — pass 4

`00640a275b26` — read the diff, the HEAD fix commit, and the surrounding source.

**Evidence basis:** finding 1 and finding 3 were verified by **RUNNING** — a `bb -cp hooks` probe calling `reviewer/parse-output`, `reviewer/mergeable?` and `reviewer/reconcile` directly. I also ran `bb test` once (`env -u PR_REVIEW_LOOP_REVIEWER`): **266 tests, 750 assertions, 0 failures, 0 errors**, which matches the HEAD commit message exactly. Findings 2 and 4 are a reading of the source. Everything ran in this throwaway worktree against temp directories; nothing touched the real ledger, `pushes.log`, or the PR.

**Closure of pass 3's blocking and coverage findings:**

- #1 (context diff keyed by sha alone) — **closed**. `context/diff-path` (context.clj:45) is `<context-dir>/<pr>-<sha>.diff`, `build!` destructures `:pr` (context.clj:61) and `review-in!` passes it (trigger.clj:461). `prune!` is per PR (context.clj:88) and the call site is updated (trigger.clj:516). The sibling — `prune!` deleting a running review's file — is closed too. `manual/decide` supplies `:pr` (manual.clj:58) and `decide-outside-a-reviewer` supplies it (trigger.clj:273), so both entry paths carry it.
- #3 (coverage: `diff-path-is-namespaced-by-sha`) — **closed**. `context_test.clj:47` now builds two PRs at one sha, asserts the paths differ and that PR 370's bytes survive PR 402's build.
- #2 (unfenced quoted verdict beats this pass's own) — **not closed for the class**, only for replies that carry the tag. Re-raised below as finding 1, with pass 3's own reproduction re-run verbatim.
- #4 (coverage: only the fenced spelling was tested) — **partly closed**; the new test covers the unfenced spelling on the tagged path only. Folded into finding 1.

Non-blocking items from pass 3 that did close, for the record: atomicfile's abandoned temps (atomicfile.clj:28, with a test that also pins the young-temp and other-path cases), prompt.clj's false `gh` sentence, SKILL.md's "no network", spec R7, the `-main` housekeeping comment, and `attempts/pusher`'s truncated docstring.

VERDICT[96e5df00]: NOT MERGEABLE — a reply with no tagged verdict still parses by the last-at-column-0 rule, so a plainly quoted prior MERGEABLE beats this pass's own NOT MERGEABLE and the ledger records a clean pass over a blocking finding

  [correctness/blocking]  1
  [correctness/followup]  2
  [coverage]              none
  [docs-accuracy]         1
  [style]                 none

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/reviewer.clj:342 — `(last (if (seq hits) hits (verdict-hits lines nil)))` falls back to the untagged last-at-column-0 rule whenever the reply contains no tagged verdict, and that fallback is pass 3's #2 unchanged. RUN, pass 3's own reproduction with the tag omitted from the reviewer's own verdict line: own `VERDICT: NOT MERGEABLE` with `[correctness/blocking] 1`, then a plain unindented, unfenced quote of a prior `VERDICT: MERGEABLE` under a "Closure of the previous pass" heading — `parse-output` returns `MERGEABLE`, `correctness/blocking 0`, `mergeable? true`, and `reconcile` confirms `MERGEABLE`. That row is what `ledger/append-pass!` writes (trigger.clj:508) and what `findings-message` turns into "you may merge this PR now". The trigger is not exotic: the core prompt's own emphatic format block, ~100 lines ahead of the "Your verdict line" section, shows the untagged spelling in both of its examples (review_core.md:42-43), and every pass before this one emitted exactly that. The only thing standing between the two is prose ("do not quote a previous verdict flush left after answering", review_core.md:38-39) — the same kind of instruction whose earlier version this defect already defeated twice. Sibling, same root: the new test `an-unfenced-quote-of-a-prior-verdict-loses-to-the-pass-tag` (test/pr_review/reviewer_test.clj:582) names the class ("the fence was one spelling, not the class") but exercises only the tagged path, and `a-reviewer-that-ignores-the-tag-is-still-parsed` (:636) certifies the fallback with a fixture that contains no quoted verdict — so nothing in the suite can observe the open half. The docstring's stated cost of closing it ("a reviewer that drops the tag must not become unreviewable... would cost a whole pass") is contradicted by the code one screen away: `MALFORMED` is handled at trigger.clj:504-508 with **no** `ledger/append-pass!` call, so it spends no cap slot and leaves the sha unreviewed and re-triggerable. Treating "tag was asked for and no tagged verdict came back" as MALFORMED costs a retry, not a pass, and closes the class.

2. [correctness/followup] claude-code-plugins/hooks/pr_review/context.clj:89 — `prune!`'s glob became `(str pr "-*.diff")`, so retention is now per PR with nothing bounding the number of PRs. TRACED: grep finds exactly one deleter of `*.diff` in the tree — this function — and it only ever runs for the PR just reviewed (trigger.clj:516), so a PR that is merged or abandoned keeps its last `context-keep` = 5 diffs in `<git-dir>/pr-review-context/` forever. Before this commit the glob was `"*.diff"` and bounded the whole directory at 5 files. The PR body reports 33 PRs reviewed in one clone; at this PR's own diff size (667 KB) that is hundreds of megabytes inside `.git` with no reclaim path. Disk only — no corruption, no wrong review — which is why this is followup. Siblings of the same "global bound split per PR, nothing bounds the PRs" shape, reported together: `ledger/trim-per-pr` (ledger.clj:154, from `05be6ac29`) keeps `max-passes-kept` rows *per PR* so the ledger file grows with PR count, and `<git-dir>/pr-review.<pr>.findings.md` (trigger.clj:329) and `<git-dir>/pr-review.<pr>.stderr` (trigger.clj:487) are written per PR and never deleted by anything. The lock file is the one that does not leak — `release!` removes it (lock.clj:190).

3. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:388 — `parse-fingerprints` reads the whole reply by design, so finding lines a reviewer quotes from a previous pass while verifying closure enter *this* row's `:fingerprints`. RUN: a reply whose own single finding is `src/new.clj:7` but which quotes the prior pass's `2. [correctness/followup] src/old.clj:99 — the old one, now fixed` unfenced yields `[src/new.clj:7:correctness/followup src/old.clj:99:correctness/followup]`. `ledger/suppressed-fingerprints` (ledger.clj:118) fires at `freq >= 2` across passes, so a follow-up raised genuinely once and quoted once is suppressed as if it had been raised twice, and the next reviewer is told not to re-raise it. Bounded to suppressible categories — blocking and coverage are filtered out by `suppressible?` — so it cannot produce a false clean. This is pass 3 #2's second half, which the tag fix does not reach because the tag discriminates verdicts, not finding lines.

4. [docs-accuracy] docs/superpowers/specs/2026-09-08-pr-review-loop.md:209 — the flow diagram still reads `context: base SHA, full diff → .git/pr-review-context/<sha>.diff`. That path has not existed since this commit renamed it to `<pr>-<sha>.diff` (context.clj:45), and the same commit edited R7 in this file without sweeping the diagram. Sibling one line up, from an earlier commit in the branch and missed by the same sweep: `:208` says `lock .git/pr-review.lock`, while `lock/lock-path` (lock.clj:35) has been `<git-dir>/pr-review.<pr>.lock` since the lock became per PR — which is the change the spec elsewhere relies on to explain why two reviews can run concurrently.

yqrashawn:
## pr-review-loop — pass 5

`945488e23227` — **NOT MERGEABLE**

Read the diff, the HEAD fix commit, and the surrounding source.

**Evidence basis:** findings 1 and 2 were verified by **RUNNING** — three `bb -cp hooks` probes calling `reviewer/parse-output`, `reviewer/reconcile`, `reviewer/mergeable?` and `reviewer/parse-warnings` directly. I also ran `bb test` once (`env -u PR_REVIEW_LOOP_REVIEWER`): **269 tests, 758 assertions, 0 failures, 0 errors**, matching the HEAD commit message exactly. Findings 3–5 are a reading of the source. Everything ran in this throwaway worktree against temp directories; nothing touched the real ledger, `pushes.log`, or the PR.

**Closure of pass 4's blocking finding**

Pass 4's blocking finding named `reviewer.clj:342`, the line

```
(last (if (seq hits) hits (verdict-hits lines nil)))
```

and the reproduction: own untagged `VERDICT: NOT MERGEABLE` with one blocking finding, then a plain unfenced quote of a prior `VERDICT: MERGEABLE`. That code is gone. RUN, pass 4's reproduction verbatim against the new `winning-verdict` (reviewer.clj:349-354): `MALFORMED`, all counts 0, `mergeable?` false. **Closed at its file:line.** `trigger.clj:530-534` confirms the cost is a retry, not a pass — no `ledger/append-pass!` on the MALFORMED branch.

Pass 4's other three also closed or were addressed: the per-PR diff/findings/stderr leak now has `sweep-abandoned-artifacts!` (trigger.clj:340-356, called at :547), and the spec's flow diagram and lock path are corrected (spec:206, :209, :245).

The class the tag exists to close is still open one spelling further down, and I reproduced it with a single formatting slip rather than two.

VERDICT[d1ef9a7b]: NOT MERGEABLE — when the reviewer's own verdict misses the tagged column-0 form and it quotes a prior pass verbatim as the prompt asks, the quoted verdict is the only column-0 one, so `winning-verdict` takes it as this pass's answer and a NOT MERGEABLE review with a blocking finding is recorded MERGEABLE

  [correctness/blocking]  1
  [correctness/followup]  2
  [coverage]              1
  [docs-accuracy]         1
  [style]                 none

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/reviewer.clj:353 — `(and tag (< 1 (count bare))) nil` rejects a reply only when it holds **two or more** column-0 verdicts. The docstring at :338-339 states the premise: "The defect needs two, one of them quoted, so one verdict cannot be it." That premise is false. The one column-0 verdict can be the quoted one, whenever the reviewer's own verdict does not match `^VERDICT\[<tag>\]:` at column 0 — which is one slip, not two, and `review_core.md:20-21` enumerates five spellings of it ("no leading whitespace, no quoting, no list marker and NOT inside a ``` fence"), each of them an observed one or that sentence would not exist. RUN, three probes, all on this HEAD: (a) own verdict carrying a list marker, `- VERDICT[d1ef9a7b]: NOT MERGEABLE`, with pass 3's verdict quoted verbatim above it under a closure heading — `parse-output` → `MERGEABLE`, `correctness/blocking 0`, `reconcile` confirms `MERGEABLE`, `mergeable?` true, `parse-warnings` `[]`; (b) own verdict indented — same result, `correctness/followup 2` picked up from the quoted block; (c) own verdict inside a fence — same. The quoted block supplies the counts too, because `vi` is the quoted verdict's index and `parse-counts` reads after it (reviewer.clj:411), so `reconcile` (:473) confirms the clean verdict instead of catching it: it only ever flips MERGEABLE→NOT MERGEABLE, never the reverse, so the verdict LINE being the quoted one is the one route it cannot cover. `parse-warnings` (:492) has no case for "a tag was asked for and none came back", so nothing in the wake says the reply was parsed by the weak rule. That row is what `ledger/append-pass!` writes (trigger.clj:537) and what `findings-message` turns into "MERGEABLE means you may merge this PR now". Reachability is not hypothesised: `prompt.clj:92-95` tells the reviewer to "quote those verbatim -- they carry their own tag, or none, and cannot be read as yours", which both invites the unfenced quote and asserts the safety that finding 1 disproves. The invariant the parse needs is provenance — the winning verdict must be one the reviewer wrote as its OWN — and the tag is the only evidence of that; `(count bare)` is a proxy for it that the quoted-verdict case defeats. Whitelisting the valid class (a tag was asked for ⇒ only a tagged verdict wins, everything else is MALFORMED) costs a retry, which trigger.clj:530 already shows is affordable, and closes the complement rather than the two listed members.

2. [coverage] claude-code-plugins/test/pr_review/reviewer_test.clj:638 — `a-reviewer-that-ignores-the-tag-is-still-parsed-when-unambiguous` states the safety property in its own `testing` string — "One verdict at column 0 cannot be the defect, which needs two, one of them quoted" — and its fixture (:642-646) is a reply with exactly one verdict which is the reviewer's own and no quoted material at all. It cannot fail for the reason it claims to check: the case that falsifies the claim (the single column-0 verdict IS the quoted one) is not constructed anywhere in the file. Its sibling `an-ambiguous-reply-that-ignores-the-tag-is-malformed` (:651) covers only the two-verdict half, and `with-no-tag-at-all-the-old-rule-stands` (:679) exercises the 1-arity, so no test in the suite can observe finding 1. Adding a fixture whose own verdict carries a list marker or a leading space, with a prior verdict quoted unfenced, fails on this HEAD — that is probe (a).

3. [docs-accuracy] claude-code-plugins/hooks/review_core.md:43 — "If you write no tagged verdict, the LAST line at column 0 starting with `VERDICT:` is taken as your answer" has been false since this commit: two untagged column-0 verdicts now discard the whole review as MALFORMED (reviewer.clj:353), and this is the copy the reviewer reads — the same commit rewrote :30-33 of this file and did not sweep ten lines down. The advice it gives is exactly inverted for the shape it now rejects: a reviewer that follows it, putting its real verdict last after a recap, loses the pass entirely. Sibling, same sweep, and the one that makes finding 1 reachable: `claude-code-plugins/hooks/pr_review/prompt.clj:94` says a quoted prior verdict carries "their own tag, or none, and cannot\nbe read as yours" — the "or none" half is false — and it asks for quotes verbatim in the same sentence that `review_core.md:30` now asks for every quotation to be fenced, so the two sections of one prompt give opposite instructions.

4. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:197 — `(cond-> {...} (oauth-token) (assoc "CLAUDE_CODE_OAUTH_TOKEN" (oauth-token)))` reads the token file twice: once as the test, once for the value. The docstring two lines up (:191-194) states the invariant this violates — "setting CLAUDE_CODE_OAUTH_TOKEN to an empty string would replace working credentials with a blank one — absent has to mean absent" — and the second read returns nil if the file is removed, emptied or briefly unreadable between the two calls, putting a nil (stringified by `:extra-env`) into the child's environment. TRACED, not raced; the window is a few filesystem calls wide, far narrower than a network round trip, and the failure is loud (the reviewer fails to authenticate, exits non-zero, and trigger.clj:530 spends no ledger slot), which is why this is followup and not blocking. `when-let` over one read is the whole fix. Related, and only a test-path inconsistency: `run!` redacts with `((or (:token-fn opts) oauth-token))` (:242) while `spawn-env` ignores `:token-fn` entirely, so a caller that overrides one gets the other — grep finds `:token-fn` set only at reviewer_test.clj:533, so nothing in production diverges.

5. [correctness/followup] claude-code-plugins/hooks/pr_review/ledger.clj:148 — the new `sweep-abandoned-artifacts!` bounds the PR count for diffs, findings and stderr (trigger.clj:349-351) but not for the ledger, which `trim-per-pr` keeps at `max-passes-kept` rows **per PR** with nothing bounding the PRs — the same shape the commit set out to close, one file short. TRACED: grep finds `trim-per-pr` as the only thing that ever drops a ledger line, and it groups by `:pr` (:157) and keeps the last N of each group, so a merged PR's rows stay for the life of the clone. Rows are one JSON line each, so at the PR body's measured 33 PRs this is kilobytes, not the hundreds of megabytes the diff files were — which is why it is followup and why I would leave it as a note rather than a fix. Worth stating only because the sweep's own docstring (trigger.clj:342-344) lists "the lock, the checkout, the diff" as the per-PR rules and the ledger is the fourth.

yqrashawn:
## pr-review-loop — pass 6

`14fed87ee39d` — **MERGEABLE**

Read the diff, the three fix commits, and the surrounding source. Verified closure of pass 5's blocking finding by running.

**Evidence basis:** findings 1 and the closure check were verified by **RUNNING** — three `bb -cp hooks:test` probes calling `reviewer/parse-output`, `reviewer/reconcile`, `reviewer/mergeable?` and `reviewer/parse-warnings` directly, plus `reviewer-test` + `prompt-test` (**54 tests, 162 assertions, 0 failures**). Findings 2–6 are a reading of the source. Everything ran in this throwaway worktree; nothing touched the real ledger, `pushes.log`, or the PR. I did not run the full `bb test` (hookinstall/prepush namespaces write git hooks).

**Closure of pass 5's blocking finding**

Pass 5 named `reviewer.clj:353`, the rule that rejected a reply only when it held two or more column-0 verdicts, and reproduced a false clean with one. That code is gone — `winning-verdict` (reviewer.clj:363-366) is now `(if tag (last (verdict-hits lines tag)) …)`. RUN, pass 5's three probes verbatim against this HEAD, each with a prior `MERGEABLE` quoted unfenced at column 0 and the reviewer's own tagged verdict carrying (a) a list marker, (b) a leading space, (c) a fence: all three return `MALFORMED`, 0 counts, `mergeable?` false, `reconcile` passes MALFORMED through, and `:reason` is populated. **Closed at its file:line.** `trigger.clj:530-534` still writes no ledger row on that branch. Pass 5's other findings: the coverage one closed (`one-column-0-verdict-can-itself-be-the-quoted-one`, reviewer_test.clj:641, is the fixture that was missing); the docs one closed (review_core.md:41-46, prompt.clj:91-97); the `spawn-env` double read closed for `spawn-env` but not for the class — see finding 2. The ledger's per-PR unbounded growth (`trim-per-pr`) is unchanged and stays on the follow-up list; I am not re-raising it as a numbered line.

Two notes on the PR body, neither filed as a finding because no repo path identifies them: "200 tests, 532 assertions" is stale (HEAD's commit message says 270/760), and "Two commits are unrelated drive-bys" undercounts — `208922923` (remove the `permission_prompt` MCP tool) is a third, and its fallout is 250 deleted lines across four `.doom.d` files. I checked that cleanup: none of the four removed elisp symbols is referenced anywhere under `.doom.d/`, and the removal is guarded by a real assertion (`gptel-claude-code-test.el:829`).

VERDICT[204c34db]: MERGEABLE — 6 follow-ups to file; pass 5's blocking finding is closed and verified by running, and nothing new meets the blocking bar

  [correctness/blocking]  none
  [correctness/followup]  2
  [coverage]              1
  [docs-accuracy]         2
  [style]                 1

1. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:477 — the reply states the same fact three times — the verdict line, the count block, and the finding lines — and only the first two are ever cross-checked, so the third is parsed, kept, and discarded as evidence. `mergeable?` reads `(zero? (get counts "correctness/blocking" 0))` while `:fingerprints` (reviewer.clj:446) already carries each finding's category, and `reconcile` (:479) only ever flips MERGEABLE to NOT MERGEABLE off the counts. RUN, two spellings that both survive the tag: (a) a reply whose count block says `[correctness/blocking] none` but whose numbered list contains `1. [correctness/blocking] src/a.clj:42` parses MERGEABLE, `mergeable?` true, `parse-warnings` `[]`, and fingerprints `["src/a.clj:42:correctness/blocking" …]` — the contradiction is in the returned map and nothing looks at it; (b) a correctly-tagged `NOT MERGEABLE` with its count block and finding, followed at the end by a restatement of the prompt's own template, `VERDICT[<this pass's tag>]: MERGEABLE -- N follow-ups to file`, parses MERGEABLE with all counts 0 and `mergeable?` true — because `last` takes the trailing echo and `parse-counts` finds no block after it. Both end with `findings-message` telling agent A "MERGEABLE means you may merge this PR now" over a listed blocking finding. Reachability for (b) is prose-guarded and asymmetric: review_core.md:45-46 documents `last` and warns only about restating the format **before** reviewing, and prompt.clj:96-97 says "Never write this tag on any line but your own final verdict" — the same kind of instruction this defect class has defeated on five previous passes, except that here the reviewer must violate an explicit instruction rather than obey one, which is why this is followup and not blocking. Both spellings close together: compare the count block against the categories already in `:fingerprints` and treat a disagreement as NOT MERGEABLE (or MALFORMED, which trigger.clj:530 shows is affordable).

2. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:248 — the "ONE read, not two" fix closed `spawn-env` and not the class: `run!` reads the token at :248 for redaction, and `default-spawn` calls `(spawn-env)` at :236, which reads it again for the child's environment. TRACED. The two values can differ, and the direction that matters is that the child gets token B while `scrub` is bound to token A, so a token the reviewer prints reaches the PR comment, the findings file and the streamed stderr unredacted — the three republication paths `redact`'s own docstring (:205-217) exists for. The window is a few filesystem calls wide, far narrower than a network round trip, which is why this is followup; it is worth stating anyway because `default-token-file` (:167) is under `~/Library/CloudStorage/Dropbox/`, a directory a sync daemon rewrites in place, which is exactly the "removed, emptied or briefly unreadable" state the new docstring names. The same two-reads shape also makes `:token-fn` half-effective: it reaches redaction at :248 and never reaches `spawn-env`, so a caller that overrides one gets the other (grep finds it set only at reviewer_test.clj:533, so nothing in production diverges today). One read in `run!`, passed into `spawn`, closes both.

3. [coverage] claude-code-plugins/test/pr_review/prompt_test.clj:24 — `base-args` never sets `:verdict-tag`, and no other test in the file does, so `prompt/build`'s entire "Your verdict line" section (prompt.clj:80-97) — the block three consecutive commits have been rewriting, and the only place the real tag is ever stated to the reviewer — is built by no test at all. Nothing asserts the prompt contains `VERDICT[<tag>]:`, and nothing asserts those two example lines are indented, which is now the sole guard against an echoed template being read as a verdict (finding 1b). The trigger side cannot cover it either: `review-opts` injects `:verdict-tag-fn (constantly test-tag)` (trigger_test.clj:98) and every `:spawn-fn` stub ignores its `prompt` argument (:109, :878, :1049), so `constantly` returns the same string however many times it is called — a regression that minted the tag twice, or dropped `:verdict-tag` from the `prompt/build` map, would leave the suite green while making **every** production review MALFORMED, since a tagless reply is now unparseable by construction. Sibling, same sweep: `the-shipped-core-prompt-is-not-mistaken-for-a-review` (reviewer_test.clj:343) states its reason as indentation — "every VERDICT line in it is an indented example, so the whole document must read as MALFORMED" — and can no longer fail for it, because review_core.md's examples are now spelled `VERDICT[<tag>]:` and the 1-arity it calls matches only `^VERDICT:`; un-indenting review_core.md:48-49 would not turn it red. I am explicitly **not** treating this as the verdict-flipping kind of coverage finding: the property that test names still holds for review_core.md by a stronger mechanism than indentation (the literal `<tag>` can never equal a minted tag), so it does not certify something false. What is uncovered is the live indentation guard one file over, in prompt.clj, which the test never claimed.

4. [docs-accuracy] docs/superpowers/specs/2026-09-08-pr-review-loop.md:257 — the Reviewer contract block still describes the pre-Bash design in four consecutive claims, three of them false: `:255` "tools are `Read,Grep,Glob`", `:257` "invariant: B cannot write, cannot run commands, cannot reach the network beyond the model API", and `:256` "stdout contains a `VERDICT:` line" (now `VERDICT[<tag>]:`, and an untagged one is refused). `:258` is false in the other direction — "no `VERDICT:` line → the trigger records `verdict: \"MALFORMED\"`" — while trigger.clj:530-534 deliberately writes **no** ledger row for MALFORMED, which is the property the last two commits' reasoning rests on. TRACED against `claude-argv` (reviewer.clj:131-158), which passes `--disallowedTools` and grants Bash, and against prompt.clj:102-106, which tells the reviewer it has Bash and `gh pr view`. Siblings, same shape and same missed sweep — pass 3 corrected R7 (`:128`) for exactly this and did not carry it through the file: `:159` "B's own Bash | irrelevant — B gets no Bash", and `:312`, in the residual-risk list, "B runs with A's full environment — `.env.local` symlinks into Dropbox, fly tokens, prod credentials. Bounded by B having no Bash and no Write" — the one place a reader goes to size the exposure, and the bound it states does not exist. C52/C54 (:102-104) already record the truth two hundred lines up, so the file contradicts itself.

5. [docs-accuracy] claude-code-plugins/hooks/review_core.md:44 — "the whole review is thrown away and **re-run**" (and prompt.clj:93, the same sentence in the generated section) promises an automatic re-run that no code performs. TRACED: `review-in!` returns `{:exit 2 :message …}` on the MALFORMED branch (trigger.clj:530-534), `review!` (:588) and `respond` (:612) have no retry, and the one path that does retry — `abandoned-candidates` (:181) — enumerates `lock/abandoned`, which a MALFORMED review is not in, because `release-quietly!` (:571) drops its lock on the way out. `reviewer.clj:357-359` gets this right ("re-triggerable"); the reviewer-facing copy does not. The consequence is in the wake: `failed-review-message` (trigger.clj:456) ends "Fix the cause and push again", which for this cause is advice agent A cannot act on — the cause is the reviewer's own output shape — and "push again" reviews a different sha, so the sha that was MALFORMED is never reviewed. The accurate remedy is `bb review-pr`, which `manual/decide` supports precisely because no ledger row was written (manual.clj:60-64), and which the message never names. Most costly on the last push before a merge, where there is no next push to carry the review.

6. [style] claude-code-plugins/hooks/pr_review/prompt.clj:97 — trailing space after the closing `)))` of the "Your verdict line" section, introduced by this commit; nothing else in the file has one.

yqrashawn:
## Review summary

Six passes. Every blocking finding is closed and verified by running, not by reading.

| pass | blocking finding | closed by |
|---|---|---|
| 1 | a quoted prior verdict inside a fence parsed as this pass's | `unfenced`, and the fence/emphasis ordering that made the first attempt inert |
| 2 | `run-review!` passed an unbound `pr`, so every checkout path was `pr-review-sci.impl.io$pr@…-<sha>` — the per-PR split did nothing and `prune-stale!` could never match | the destructuring, plus a stub that records its `pr` instead of discarding it |
| 3 | the context diff was keyed by sha alone, so two open PRs at one sha overwrote each other's diff mid-review and findings landed on the wrong PR | `<pr>-<sha>.diff`, and `prune!` scoped per PR |
| 3 | an **unfenced** quote of a prior verdict beat this pass's own | a per-pass verdict tag |
| 4 | the untagged fallback was the same defect by another route | untagged ⇒ parse only if unambiguous |
| 5 | one column-0 verdict can itself be the quoted one — a single formatting slip, not two | a tag was asked for ⇒ only a tagged verdict wins, everything else MALFORMED |

The verdict-parse class took four passes because each fix replaced one proxy for the real property with another. The property is **provenance** — the winning verdict must be one the reviewer wrote as its own — and fences, position, and the count of column-0 verdicts are all proxies that a quotation defeats. The per-pass tag is the only direct evidence, so it is now required: a reply without one is MALFORMED, which costs a retry and no ledger row.

Also fixed across the six passes: a fixed `.tmp` sibling that could rename a partial `pre-push` into place (a non-zero pre-push aborts the user's push), temps abandoned by a kill, `:pushed-by` never set on the one path that needs it, per-PR retention with nothing bounding the PRs, `spawn-env` reading the token file twice, and a dozen doc claims that contradicted the code.

### Not fixed here

Pass 6 is MERGEABLE with 6 non-blocking findings. Per the loop's own termination rule these do not get pushed to this PR:

- **followup** — the count block and the finding lines are never cross-checked, so a reply whose list contains a blocking finding but whose count block says `none` parses MERGEABLE; and `run!` reads the OAuth token separately from `spawn-env`, so a token rotated between the two reads reaches the PR comment unredacted. Both are worth a follow-up PR.
- **coverage** — `prompt_test` never sets `:verdict-tag`, so the prompt section that states the tag is built by no test. Real gap, not a false-safety signal.
- **docs-accuracy** — the spec's Reviewer contract block still describes the pre-Bash design (C52/C54 already record the truth 200 lines up), and review_core.md promises an automatic re-run that no code performs; the accurate remedy is `bb review-pr`.
- **style** — one trailing space.

### CodeRabbit

Its findings overlapped this loop's substantially and were valid where they overlapped: the `pushes.log` lock, `:pushed-by`/`:pr-url`, the review_core CI claim, the ledger's global trim, PID reuse, chmod-after-move, the checkout collision, and the activation cache. It under-counted only the `.doom.d` fallout — 6 dead tests where there were 14, because `--format-tool-input-summary` had been removed too.

PR #156, opened 2026-09-15
chore(gpg): raise agent cache TTL from 7 to 60 days

yqrashawn:
Both `default-cache-ttl` and `max-cache-ttl` were 604800 seconds (7 days), so the GPG passphrase had to be re-entered weekly. Raised both to 5184000 seconds (60 days) in the generated `~/.gnupg/gpg-agent.conf`.

The SSH-specific options (`default-cache-ttl-ssh` / `max-cache-ttl-ssh`) are not set here and are left unchanged.

Requires a rebuild plus `gpgconf --reload gpg-agent` to take effect.

## Summary by CodeRabbit

* **Enhancements**
  * Increased the GPG agent cache duration, allowing credentials to remain cached for longer periods.


yqrashawn:
## pr-review-loop — pass 1

`60d65a8e6d62` — **MERGEABLE**

Read diff, `gpg-agent` man page from nix store, and the gpg call sites in the repo.

**Method:** all findings are a reading of source plus the packaged `gpg-agent(1)` man page (`/nix/store/…gnupg-2.4.8/share/man/man1/gpg-agent.1.gz`). Nothing was run against a live agent, no test suite run.

**Arithmetic checks out:** 604800 = 7 d, 5184000 = 60 d exactly. Commit message accurate. `default-cache-ttl`/`max-cache-ttl` take unsigned seconds; 5184000 far below any width limit — no overflow. `enable-ssh-support` was dropped in 32512f0da, so the untouched `*-cache-ttl-ssh` defaults are irrelevant here — not a finding.

VERDICT[39e6e6f6]: MERGEABLE — 3 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none
  [docs-accuracy]         none
  [style]                 1 finding

1. [correctness/followup] .doom.d/better-default.el:765 — every Emacs start flushes the cache this PR is trying to lengthen, so effective TTL = min(60 d, time until next Emacs restart), not 60 d. TRACED: `pinentry-start` carries `;;;###autoload` (`~/.emacs.d/.local/straight/repos/pinentry/pinentry.el:157`), so the `doom-after-init` hook at `better-default.el:759` loads the package, which runs the `:config` body — `gpg-connect-agent reloadagent /bye`. Man page, SIGHUP section: `This signal flushes all cached passphrases and … the configuration file is read again`; RELOADAGENT is SIGHUP. Sibling sweep: the repo has exactly two `gpg-connect-agent` call sites — `zsh.nix:181` sends only `updatestartuptty`, which does not flush, so this is the single offender; `better-default.el:767` is a commented-out `gpgconf --reload gpg-agent` with the same effect if re-enabled. Reachability: ordinary, every Emacs restart. Pre-existing line, not a regression from this diff — hence followup, not blocking.
2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:114 — nothing reloads the agent when this file changes, so a running `gpg-agent` keeps 604800 after `darwin-rebuild switch` until it is SIGHUP'd or the process dies. TRACED: no `onChange` on the `gpg-agent-conf` `home.file` entry (lines 98–122), and `grep -rn "onChange|reloadagent|gpgconf --kill" --include="*.nix"` returns nothing across the repo; `home-manager/activations/default.nix` touches gpg only for `~/.authinfo.gpg` (line 56). Reload is safe and sufficient here — the man page lists both `default-cache-ttl` and `max-cache-ttl` among the SIGHUP-honored options. Adjacent precedent the author already set: b7bdbaabf ("make a rotated token restart the agent"). Coupled to finding 1: the only reload path that currently exists is the one that also wipes the cache.
3. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:114 — `default-cache-ttl` is inert when set equal to `max-cache-ttl`. Per the man page, `default-cache-ttl` is the idle timer (reset on each access) and `max-cache-ttl` the absolute cap from creation; equal values mean the sliding reset can never extend anything, so the real policy is "60 days from creation". If a shorter idle timeout was wanted, `default-cache-ttl` is the knob and it is currently pinned.

Not a finding, stated once: `max-cache-ttl 5184000` moves the practical security boundary from "TTL" to "agent process lifetime" — for 60 days any process running as the user can drive the secret key through the agent socket with no prompt, and `ignore-cache-for-signing` is unset so commit signing rides the same cache. That is the explicit intent of the commit, so it is a tradeoff, not a defect.

PR #157, opened 2026-09-16
fix(gpg): stop Emacs flushing the agent cache, reload it on config change

yqrashawn:
Follow-up to #156, which raised the gpg-agent cache TTL to 60 days. Two things made that TTL not actually hold.

1. `.doom.d/better-default.el` ran `gpg-connect-agent reloadagent /bye` on every Emacs start. RELOADAGENT is SIGHUP, which flushes all cached passphrases, so the effective TTL was the time until the next Emacs restart rather than 60 days. The `updatestartuptty` call on the following line is what the pinentry setup actually needs, and it does not flush, so only the `reloadagent` line is removed.

2. Nothing reloaded the agent when `gpg-agent.conf` itself changed, so a running agent kept the old TTL after a rebuild until it was signalled or died. Added an `onChange` hook on the `home.file` entry that sends `reloadagent` — the config-change case is exactly when flushing the cache is correct.

Both findings came from the review on #156. The style finding there (`default-cache-ttl` being inert when equal to `max-cache-ttl`) is intentional and not changed here.


yqrashawn:
## pr-review-loop — pass 1

`f676e72a5f41` — **MERGEABLE**

VERDICT[f99d84f8]: MERGEABLE -- 4 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  4 findings
  [coverage]              none
  [docs-accuracy]         none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:122 — unguarded `onChange` can abort the whole home-manager activation. TRACED: home-manager's activation script runs under `set -eu` / `set -o pipefail` (`/nix/store/qndj4ringiyvm4jdjh81cwlja8idfv64-source/modules/home-environment.nix:807`) and interpolates `onChange` bare, with no `run` wrapper and no `|| true` (`.../modules/files.nix:283`). RAN: with no agent reachable and autostart failing, `gpg-connect-agent reloadagent /bye` exits **1** ("failed to start gpg-agent ... General error"); with `--no-autostart` and no agent it exits 0, so only the failed-start case bites. Trigger: conf content changed AND gpg-agent not running AND fails to start (bad `~/.gnupg` ownership, or an unknown option in the conf — gpg-agent exits 2 on those, so a typo in the `text` block at :113-119 would turn a cosmetic mistake into a failed rebuild). `onFilesChange` is `entryAfter ["linkGeneration"]`, which sorts alongside `installPackages` and the repo's own `home.activation.defaultSetup` (`modules/yqrashawn/home-manager/activations/default.nix:9`) — those can be skipped. Upstream guards this: `modules/services/mako.nix:159` uses `|| true`, `programs/kitty.nix:347` the same, hyprland gates on a directory check. Also `>/dev/null` leaves stderr unredirected, so the failure path prints gpg noise into every rebuild (compare `modules/yqrashawn/home-manager/cli/zsh.nix:181`, which uses `2>&1`).

2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:118 — the 60-day TTL is still capped, now by nixpkgs churn instead of Emacs restarts. TRACED: the conf text embeds `${pinentry-auto}`, whose store path is derived from `pkgs.pinentry-curses` and `pkgs.pinentry_mac` (:102-106). `checkFilesChanged` compares the *content* of the new store file against `$HOME/.gnupg/gpg-agent.conf` with `cmp` (`.../modules/files.nix:238-241`), so any pinentry bump from a routine `nix flake update` counts as a config change, fires `reloadagent`, and RELOADAGENT flushes **all** cached passphrases (`gpg-agent(1)`, SIGHUP: "This signal flushes all cached passphrases"). So the effective TTL becomes the flake-update cadence for pinentry. Still a large improvement over per-Emacs-start, but the PR body's "the config-change case is exactly when flushing the cache is correct" does not hold for pure store-path churn — no semantic option changed. gpg-agent offers no reload-without-flush, so the fix is to gate the hook on the options that matter (e.g. reload only when the TTL/pinentry *values* change) rather than on file content.

3. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:114 — the change does not apply the new TTL to the agent that is running today. TRACED: the conf text is byte-identical to the base commit (the diff only adds :122), so `changedFiles[".gnupg/gpg-agent.conf"]=0` on the activation that merges this PR and the hook does not fire. A gpg-agent that has been alive since before #156 keeps its old `default-cache-ttl` until it restarts, and this PR removes the Emacs `reloadagent` that previously applied it — nothing else in the repo signals the agent (grep for `reloadagent|gpgconf|killagent` across the tree hits only `.doom.d/better-default.el`, `dotfiles/default.nix`, `cli/zsh.nix`; nix-darwin's job at `modules/yqrashawn/darwin/core.nix:109` only runs `gpg-connect-agent /bye`, which does not reload). One-time and self-healing on the next agent restart, so bounded — worth a `gpgconf --reload gpg-agent` once by hand.

4. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:117 — `reloadagent` silently cannot apply every option in this file. TRACED from `gpg-agent(1)`: the SIGHUP honored list is `quiet, verbose, debug*, no-grab, pinentry-program, pinentry-invisible-char, default-cache-ttl, max-cache-ttl, ignore-cache-for-signing, s2k-count, no-allow-external-cache, allow-emacs-pinentry, no-allow-mark-trusted, disable-scdaemon, disable-check-own-socket`. `allow-loopback-pinentry` (:117) is not on it, and Emacs depends on it via `epg-pinentry-mode 'loopback` (`.doom.d/better-default.el:761`). Today's values are already live, so nothing is broken now; the trigger is a future edit to that option, which the hook will report as handled while the agent ignores it. A restart (`gpgconf --kill gpg-agent`) is the only way to apply it.

5. [style] .doom.d/better-default.el:765 — the surviving call is a synchronous `shell-command` on the `doom-after-init` path, and `>/dev/null` covers stdout only, so agent stderr pops a `*Shell Command Output*` window at startup. `modules/yqrashawn/home-manager/cli/zsh.nix:181` is the established convention in this repo (`>/dev/null 2>&1`), as is nix-darwin's own `--quiet ... > /dev/null 2>&1`. Pre-existing line, adjacent to the deletion.

Method: I RAN only `gpg-connect-agent` against two throwaway `--homedir` directories in the scratchpad to establish its exit codes (finding 1); no agent was started and the user's running agent was not contacted. Everything else is a reading of the diff, `.doom.d/better-default.el`, the repo's Nix modules, the pinned home-manager and nix-darwin module sources in the store, and `gpg-agent(1)`.

PR #158, opened 2026-09-16
fix(gpg): guard the agent onChange hook and stop store churn flushing the cache

yqrashawn:
Follow-up to #157, addressing three findings from its review.

**Unguarded hook could abort activation.** home-manager runs the activation script under `set -eu` and interpolates `onChange` bare, so a non-zero exit from the hook fails the rebuild. `gpg-connect-agent` exits 1 when no agent is reachable and autostart fails — for example when an unknown option in this very file makes gpg-agent refuse to start, turning a typo into a broken rebuild. The hook now ends in `|| true` and redirects stderr as well as stdout, matching `cli/zsh.nix:181` and upstream's own guarded hooks.

**Store-path churn counted as a config change.** The conf text embedded the `pinentry-auto` store path, and home-manager decides `changedFiles` by comparing file content. Any pinentry bump from a routine `nix flake update` therefore fired the hook and flushed every cached passphrase, capping the 60-day TTL at the flake-update cadence. The script is now installed at a stable `~/.gnupg/pinentry-auto` and the conf points there, so the conf changes only when an actual option value changes.

**reloadagent does not apply every option here.** Per `gpg-agent(1)`, the SIGHUP list does not include `allow-loopback-pinentry`, which Emacs depends on via `epg-pinentry-mode 'loopback`. The hook now runs `gpgconf --kill gpg-agent`; the agent is autostarted on next use and comes back with the full config applied.

Not addressed here, since it is a one-time manual step rather than a code change: the agent running right now still holds the pre-#156 TTL and needs `gpgconf --kill gpg-agent` once by hand after the next rebuild.


yqrashawn:
## pr-review-loop — pass 1

`4c00d71a6cf6` — **MERGEABLE**

Read the diff, then traced the mechanics through home-manager's `modules/files.nix` and nix-darwin's `modules/programs/gnupg.nix` in `/nix/store`.

**Verified by RUNNING:** `nix-instantiate --parse` on the changed file (syntax OK). Everything else below is a reading of the source (the diff, home-manager `files.nix` / `file-type.nix` / `programs/gpg.nix`, nix-darwin `programs/gnupg.nix`, and this repo's `darwin/core.nix`, `cli/gpg.nix`, `cli/zsh.nix`).

**What I confirmed as correct**, since the PR's central claims are checkable:
- Churn claim (traced): `files.nix:266` compares `cmp --quiet <sourceStorePath> $HOME/<target>` — content, following symlinks. Removing the pinentry store path from the conf text does stop pinentry bumps from firing the hook.
- Guard claim (traced): `home-environment.nix:807-808` sets `set -eu` + `set -o pipefail`, and `files.nix:283` interpolates `${v.onChange}` bare. `|| true` is a real fix.
- SIGHUP claim (traced against gpg-agent(1) option list): `allow-loopback-pinentry` is genuinely absent from the reload list.
- Executable bit survives: `executable = null` → `inherit` → `ln -s` (`files.nix:371`, `file-type.nix:78`), so `~/.gnupg/pinentry-auto` keeps `writeShellScript`'s 0555.
- `~/.gnupg` is already 0700 before linking — home-manager's `createGpgHomedir` is `entryBetween ["linkGeneration"] ["writeBoundary"]`, so it runs first. No perms regression from adding a second file there.

VERDICT[98d28922]: MERGEABLE — 5 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:125 — `gpgconf --kill gpg-agent` tears down the agent's ssh socket, which `darwin/core.nix:111` (`programs.gnupg.agent.enableSSHSupport = true`) binds `SSH_AUTH_SOCK` to; `reloadagent` did not. Traced: nix-darwin's `modules/programs/gnupg.nix` defines `launchd.user.agents.gnupg-agent` as `gpg-connect-agent /bye` with `KeepAlive.SuccessfulExit = false` — that command exits 0, so launchd never relaunches it after the kill. Already-open shells keep a stale `SSH_AUTH_SOCK` pointing at a removed socket and lose agent keys until some gpg command autostarts the agent or a new login shell runs `environment.extraInit`'s `updatestartuptty`. Bounded and self-healing, hence followup. Trigger: any activation where gpg-agent.conf content changes — including this PR's own first activation (see finding 3).
2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:103 — INFERRED, not traced, and pre-existing (these lines are relocated verbatim, not authored here): gpg-agent speaks Assuan to pinentry over pipes on stdin/stdout, so `[ -t 0 ]` is false in every agent-spawned invocation and the `elif` branch can never select curses. Curses is reachable only via the `PINENTRY_USER_DATA` branch, which `cli/zsh.nix:176` sets solely under `$SSH_CONNECTION`. A headless session that is not an SSH session therefore gets pinentry-mac. Worth confirming since the PR re-indents and re-homes exactly this script.
3. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:118 — the PR body defers a step it does not need to: "the agent running right now still holds the pre-#156 TTL and needs `gpgconf --kill gpg-agent` once by hand after the next rebuild." Traced: this line changes the conf's `pinentry-program` value from a store path to `$HOME/.gnupg/pinentry-auto`, so `files.nix:266`'s content compare yields `changedFiles=1` and `files.nix:283` runs the new kill hook during that very activation. The manual step is already performed by the diff.
4. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:123 — "The agent is autostarted on next use" is unqualified but only holds for gpg CLI use. Per finding 1, ssh use through `SSH_AUTH_SOCK` does not autostart the agent, and this config enables SSH support. The comment reads as a completeness guarantee it does not have.
5. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:108 — `.gnupg` is now hardcoded in two places (target here, and the interpolated path at :118) while `cli/gpg.nix` enables `programs.gpg`, whose `homedir` option (default `${config.home.homeDirectory}/.gnupg`, home-manager `programs/gpg.nix:241`) is the module-level source of truth and also drives `home.sessionVariables.GNUPGHOME`. `config.programs.gpg.homedir` would keep them in sync; the duplication is new even though the hardcoding is not.

PR #159, opened 2026-09-16
fix(gpg): restart the agent after killing it, drop the dead pinentry branch

yqrashawn:
Follow-up to #158, addressing its two correctness findings.

**The kill left SSH_AUTH_SOCK dangling.** `programs.gnupg.agent.enableSSHSupport` is on (`darwin/core.nix`), and nix-darwin's launchd job is `gpg-connect-agent /bye` with `KeepAlive.SuccessfulExit = false`, so launchd does not bring the agent back after `gpgconf --kill`. Already-open shells kept a stale `SSH_AUTH_SOCK` until some gpg command happened to autostart the agent; ssh use alone does not. The hook now starts the agent again immediately, so the sockets are recreated at their fixed paths within the same activation. Both commands stay guarded with `|| true`.

**The curses branch was unreachable.** gpg-agent speaks Assuan to pinentry over pipes on stdin/stdout, so `[ -t 0 ]` is false in every agent-spawned invocation and that branch could never select curses. Curses is still reached through the `PINENTRY_USER_DATA` check, which `cli/zsh.nix` sets under `$SSH_CONNECTION` — the actual headless case. Removing the dead branch leaves behavior unchanged and stops the script implying a fallback it does not have.

The docs-accuracy and style findings on #158 are dropped: the deferred manual step it flagged is moot now that the hook runs on that activation, and `.gnupg` being spelled out rather than read from `config.programs.gpg.homedir` matches how the rest of this file names its targets.


yqrashawn:
## pr-review-loop — pass 1

`72fa123bd7e5` — **MERGEABLE**

Verified by RUNNING (on the live machine, read-only): `gpgconf --list-options gpg-agent`, `gpg-agent --gpgconf-list`, `man gpg-agent`/`gpgconf`, `stat` on `~/.gnupg/S.gpg-agent*`, `ps -p 1459`, plus reading nix-darwin's `programs/gnupg.nix` and home-manager's `modules/files.nix` out of the store. Everything else is a reading of the diff and its neighbours.

Premises that check out, so they are not findings: the launchd job really is `gpg-connect-agent /bye` with `KeepAlive.SuccessfulExit=false` and `RunAtLoad` (`~/Library/LaunchAgents/org.nixos.gnupg-agent.plist`), so nothing restarts the agent after the kill; the SIGHUP claim on line 120-121 matches the man page's honored-option list (`allow-loopback-pinentry` is absent, `allow-emacs-pinentry` present); and the ssh socket claim on line 122-123 holds because gnupg 2.4 creates `S.gpg-agent.ssh` on Unix regardless of `enable-ssh-support` (which this conf does not set) — I checked that specifically because the conf omits the option.

VERDICT[8e06153d]: MERGEABLE — 2 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none
  [docs-accuracy]         none
  [style]                 none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:127 — `gpg-connect-agent /bye` runs with no wait after `gpgconf --kill`, and `gpgconf --kill` returns when the agent acknowledges KILLAGENT, not when it has exited and unlinked its sockets; the new agent can therefore come up against the dying one. Live corroboration (RAN): agent pid 1459, started `Tue Sep 15 15:15:13 2026`, is still running, and `S.gpg-agent.extra`, `S.gpg-agent.browser`, `S.gpg-agent.ssh` all still carry mtime `2026-09-15 15:15:13`, while `S.gpg-agent` at the same path was re-bound at `2026-09-16 08:13:56` — the main socket file was replaced underneath a live agent, leaving two agents on one homedir with `SSH_AUTH_SOCK`'s socket owned by the old one, which is the inverse of what lines 122-123 promise. That the 08:13:56 event came from this exact command pair is INFERRED, not traced: the deployed conf is still the pre-#156 7-day one (`default-cache-ttl 604800` on disk vs `5184000` in the tree), so home-manager's hook itself has not run yet (`modules/files.nix` gates `onChange` on a content `cmp` of the target — traced). The failure-to-start half of the race is also inferred from gnupg's shutdown semantics, not from source available here. Base had the same race with a much wider gap (kill, then some later shell's `updatestartuptty` autostarts); running the start immediately narrows it to the worst moment. Both commands are `>/dev/null 2>&1 || true`, so whichever way it goes wrong is silent. A poll for the socket to disappear (or a short retry loop) before the start would close it.

2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:100 — with the `[ -t 0 ]` branch gone, `PINENTRY_USER_DATA` is the script's only headless signal, and it is exported solely from `programs.zsh.initContent`, which lands in `.zshrc` (interactive shells only — traced at `modules/yqrashawn/home-manager/cli/zsh.nix:174-183`). A non-interactive remote command such as `ssh host 'gpg -d f.gpg'`, or any non-zsh caller, gets `pinentry-mac`, which needs a GUI login session. The deletion itself is correct — the agent spawns pinentry as an Assuan pipe server so stdin is never a tty (INFERRED from the pinentry protocol, consistent with the author's reasoning) — and behaviour is unchanged from base; the point is that the hole the dead branch appeared to cover is still open, and nothing in the script now marks it.

PR #160, opened 2026-09-16
fix(gpg): wait for the old agent to exit, export PINENTRY_USER_DATA from zshenv

yqrashawn:
Follow-up to #159, addressing its two correctness findings.

**Race between the kill and the restart.** `gpgconf --kill` returns when the agent acknowledges KILLAGENT, not when it has exited and unlinked its sockets, so starting a new agent immediately could leave two agents on one homedir with `SSH_AUTH_SOCK`'s socket still owned by the dying one — the inverse of what the hook promises. The hook now polls `gpg-connect-agent --no-autostart /bye` until it fails (no agent reachable) before starting the replacement, capped at 5 seconds so a stuck agent cannot hang activation. Everything stays guarded.

**PINENTRY_USER_DATA was interactive-only.** It was exported from `programs.zsh.initContent`, which lands in `.zshrc`, so `ssh host 'gpg -d file.gpg'` — non-interactive, never sources `.zshrc` — fell through to `pinentry-mac` and its need for a GUI login session. The export moves to `envExtra` (`.zshenv`), which non-interactive shells do read. The `updatestartuptty` call stays in `.zshrc`, since it only makes sense for a shell that has a TTY.

This closes the hole that #159's deleted `[ -t 0 ]` branch appeared to cover but never did.


yqrashawn:
## pr-review-loop — pass 1

`6cef1ff27091` — **MERGEABLE**

VERDICT[27edeb97]: MERGEABLE — 7 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  4 findings
  [coverage]              none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:132 — the poll tests *reachability*, not exit-and-unlink, so the socket-steal race is narrowed rather than closed. `gpg-connect-agent --no-autostart /bye` failing proves only that nothing answers on the socket; the comment at :127-129 promises "exited and unlinked its sockets". `gpg-agent(1)` (gnupg 2.4.8 man page, SIGTERM) documents "Shuts down the process but waits until all current requests are fulfilled" — so with a request in flight at kill time the agent lingers, and once it stops accepting new connections the probe breaks early, :136 autostarts a replacement that rebinds `S.gpg-agent`/`S.gpg-agent.ssh`, and the dying agent's cleanup then unlinks the *new* agent's sockets — the dangling `SSH_AUTH_SOCK` that :122-123 exists to prevent. TRACED: :132 only tests connectivity and :136 is unconditional. INFERRED (not verified against gnupg source): that gpg-agent closes its listener before it unlinks. Trigger: a `home-manager switch` concurrent with in-flight gpg/ssh-agent use. A pid-based wait (read the agent pid via `gpgconf --list-dirs`, then `kill -0` until it is gone) would test the thing the comment claims.

2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:131 — the 50-iteration cap does not bound the case it was written for. `gpg-connect-agent` has no connect or read timeout (no such option in `gpg-connect-agent(1)`), so an agent that accepts the connection but never replies — the "stuck agent" — blocks inside the probe and the loop counter never advances; activation hangs indefinitely. `gpgconf --kill` at :126 is the same shape and equally untimed (sibling, reported here). The cap only fires when the agent is responsive, i.e. the case that did not need it. A `timeout`-wrapped probe, or the pid check from finding 1, bounds it for real.

3. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:136 — timeout exhaustion silently skips the restart the hook exists to perform. If the loop runs all 50 iterations the old agent is by construction still reachable, so `gpg-connect-agent /bye` connects to *it* instead of starting a replacement. The new `gpg-agent.conf` is then never applied — exactly the stale-config state :120-121 says a full restart is required to avoid — and nothing is printed, so the user has no signal. There is no post-loop check and no else branch: the loop's outcome is discarded.

4. [correctness/followup] modules/yqrashawn/home-manager/cli/zsh.nix:24 — `PINENTRY_USER_DATA` is the sole selector for `pinentry-auto` (dotfiles/default.nix:100), and it is now exported from exactly one place, zsh's `.zshenv`. `programs.bash.enable = true` (zsh.nix:119) with `initExtra`/`profileExtra` that never set it; `.zshenv` is also skipped for `zsh -f`, `NO_RCS`, and zsh invoked as `sh`. Latent today because the login shell is `pkgs.zsh` (modules/yqrashawn/common.nix:52) — trigger is a login-shell change or a forced command run under bash. The selector script itself could decide from a tty/`DISPLAY` probe instead of depending on every shell rc to export the variable.

5. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:128 — two false claims about the loop. The comment says it will "wait for it to actually go away", i.e. "exited and unlinked its sockets"; the code waits only until nothing answers (finding 1). The PR body says "capped at 5 seconds so a stuck agent cannot hang activation"; the cap bounds the sleeps, not the probes, and does not apply to a stuck agent (finding 2).

6. [docs-accuracy] modules/yqrashawn/home-manager/cli/zsh.nix:22 — the PR body's "This closes the hole" overclaims for its own headline example. `ssh host 'gpg -d file.gpg'` without `-t` gets no pty at all, so selecting pinentry-curses does not give it a prompt — it changes which pinentry fails. `GPG_TTY` is also still set only in `initContent`/`.zshrc` (zsh.nix:178), which that shell does not read. What the change does fix, and what the comment could say instead: `ssh -t host '...'` and any case where the passphrase is already cached. TRACED: the `.zshenv` placement (confirmed against the generated `~/.zshenv`, which carries `envExtra` verbatim) and the `GPG_TTY` split. INFERRED: that pinentry-curses cannot prompt without a terminal.

7. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:130 — `waited` is an unscoped variable in a shared namespace. home-manager inlines `${v.onChange}` straight into the activation script (modules/files.nix:283, `set -eu -o pipefail` at modules/home-environment.nix:807) with no subshell and no `local`, so the name persists for every hook that runs after. No collision today — this is the only `onChange` in the repo — but a prefixed name costs nothing.

Evidence basis: every finding above is a reading of the source. I ran nothing that exercises the changed code — the one probe I attempted (`gpg-connect-agent --no-autostart /bye` against a scratch `GNUPGHOME`) was denied, so the `--no-autostart` exit status is assumed, not measured; if it were 0 on a missing agent the loop would degrade to a fixed 5s sleep rather than misbehave, which is why I did not raise it. What I did execute was read-only: `grep`/`sed` over the worktree, `cat` of the generated `~/.zshenv`, reads of the home-manager module source in `/nix/store`, and `gzcat` of the gnupg 2.4.8 `gpg-agent`/`gpg-connect-agent`/`gpgconf` man pages.

PR #161, opened 2026-09-16
fix(gpg): bound the agent restart hook and wait on the pid

yqrashawn:
Follow-up to #160, which added a poll loop that did not do what its comment claimed. Rewritten rather than patched.

- **Waits on the pid, not on reachability.** The agent stops answering on its socket before it unlinks it, so `#160`'s connectivity probe could break early and let the replacement bind sockets that the still-dying agent then removed. The hook now reads the agent's pid up front (`gpg-connect-agent --no-autostart 'GETINFO pid' /bye` — verified against the live agent, returns the pid on a `D` line) and waits with `kill -0` until that process is actually gone.
- **Every step is really bounded.** `gpg-connect-agent` has no connect or read timeout, so a wedged agent blocked inside the old probe and the iteration cap never advanced — activation could hang indefinitely. All three agent calls are now wrapped in `timeout 5`, and the wait loop's condition is a `kill -0` that cannot block.
- **Failure is no longer silent.** If the restart does not succeed the hook prints a warning naming the manual fix, instead of quietly leaving the old agent running with the old config — the exact stale state the restart exists to prevent.
- Variables are prefixed (`gpgAgentPid`, `gpgAgentWaited`) since home-manager inlines `onChange` into the shared activation script with no subshell.

Dropping the other two findings from #160, with reasons. The bash/`zsh -f` gap for `PINENTRY_USER_DATA` is latent only under a login-shell change, and the suggested alternative — having the selector script probe for a tty — cannot work, since gpg-agent speaks Assuan to pinentry over pipes and stdin is never a tty there; that was #159's dead branch. And the doc nit about `ssh host 'gpg -d'` without `-t` is right that curses cannot prompt without a pty, but selecting curses is still the correct choice for every SSH case that can prompt.


yqrashawn:
## pr-review-loop — pass 1

`bb3c9b9db4a6` — **MERGEABLE**

Verified by **running**: the Nix file parses (`nix-instantiate --parse`), the control flow of the hook reproduced in an isolated harness with stubbed input (no agent contacted) — empty pid ⇒ zero wait, cap-exhaustion ⇒ fall through to restart, and `set -euo pipefail` safety of every construct; `GETINFO pid` is a real, documented agent command (`strings` on the gpg-agent binary: `pid - Return the process id of the server.`); home-manager inlines `onChange` with no subshell and the activation script sets `set -eu; set -o pipefail` (read from the pinned home-manager source in the store, `modules/files.nix:283`, `modules/home-environment.nix:807`). Everything else is a **reading** of the source.

I also checked the premise of the retained comment on line 122–123: `programs.gnupg.agent.enableSSHSupport = true` is set (`modules/yqrashawn/darwin/core.nix:111`), and `~/.gnupg/S.gpg-agent.ssh` has an mtime equal to the running agent's start time, so the socket genuinely is created and genuinely does disappear across a kill. That comment holds; I am not reporting it.

VERDICT[994dc6d5]: MERGEABLE — 3 follow-ups to file, plus 1 docs and 2 style

  [correctness/blocking]  none
  [coverage]              none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:135 — `if [ -n "$gpgAgentPid" ]` makes the wait conditional on the probe having succeeded, so any path where line 131's probe fails against a *live* agent skips the wait entirely and restarts with zero delay — exactly the unlink race the hook exists to close, with no warning. Traced: `timeout 5` expiring on a wedged agent, an agent in `shutdown_pending` (stops accepting new connects, sockets still bound), or an agent owned elsewhere all yield an empty pid and take the no-wait branch. Verified by running the stubbed harness: empty pid prints `NO WAIT AT ALL`. Trigger is a degraded agent, not ordinary use. A bounded unconditional fallback sleep when the pid is unknown closes it.
2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:142 — the final `gpg-connect-agent /bye` proves only that *an* agent answers, not that a *new* one started, so it cannot detect the case it is there for. When the 50-iteration cap is exhausted with the pid still alive (loop at :137 exits on the counter, not on death), control falls straight through to :142; if the old agent is still answering, the `||` branch never fires and the user is told nothing while the config change silently did not take effect. Verified by running: the harness reports `loop exited, waited=3 (cap reached => agent still alive)` and proceeds. This also makes the PR body's third bullet — "If the restart does not succeed the hook prints a warning … instead of quietly leaving the old agent running with the old config" — false for that failure mode. Plausible trigger I traced but could not exercise: an open pinentry prompt keeps a connection active, and gpg-agent will not exit while `active_connections != 0`, so a `switch` during a passphrase prompt exhausts the 5 s cap. Re-reading `GETINFO pid` after :142 and comparing against `$gpgAgentPid` would make the check say what it claims.
3. [correctness/followup] modules/yqrashawn/home-manager/cli/zsh.nix:188 — sibling of the defect the PR's second bullet names. The PR wraps all three agent calls in `timeout 5` because "`gpg-connect-agent` has no connect or read timeout, so a wedged agent blocked … activation could hang indefinitely"; the identical unbounded call `gpg-connect-agent updatestartuptty /bye` sits on the SSH interactive-shell path with no `timeout`, where a wedged agent hangs every new remote shell — a wider blast radius than a hook that runs only on config change. Not introduced by this diff; reported here because the root cause is the one this diff fixes in one of its two locations. (The equivalent call in nix-darwin's `environment.extraInit` is upstream and out of scope.)
4. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:143 — the warning fires when the agent *did not restart*, but tells the user to `gpgconf --kill gpg-agent`, which is the step that already ran and does not start anything. The recovery for the stated condition is a kill followed by a start (`gpgconf --kill gpg-agent && gpg-connect-agent /bye`); as written the advice leaves the reader one step short.
5. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:133 — `pkgs.gawk` is pulled into the activation closure to extract one field from one line. `| { read -r tag pid _ && [ "$tag" = D ] && echo "$pid"; }` is equivalent with no added dependency, and asserts the `D` tag rather than pattern-matching it. (#subtract)
6. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:137 — `kill` is the only unqualified command in a hook where every other command is a store path. It is correct and deliberately so (bash builtin, therefore PATH-independent and fork-free, which matters under home-manager's `emptyActivationPath`), but nothing says that, so the next reader's instinct will be to "fix" it to `${pkgs.coreutils}/bin/kill`. One clause of comment prevents that. (#negative-space)

Two orderings I enumerated that do **not** break: pid reuse between the old agent's exit and the `kill -0` (worst case is waiting the full 5 s, or an early exit on `EPERM`, then restarting anyway — benign either way), and an agent autostarted by a concurrent `gpg` between the probe and the kill (the kill takes it, `:142` starts a fresh one). (#temporal)

No `[coverage]` finding: `flake.nix:310` builds the darwin toplevel only, so no test asserts anything about this hook's behaviour — there is no test making a false safety claim.

PR #162, opened 2026-09-16
fix(gpg): verify the agent actually restarted, bound updatestartuptty

yqrashawn:
Follow-up to #161, addressing its three correctness findings.

- **A failed pid probe no longer skips the wait.** The wait was conditional on the probe succeeding, so a wedged agent — or one already refusing connections with its sockets still bound — produced an empty pid and a zero-delay restart, which is the unlink race the hook exists to close. The empty-pid branch now sleeps a second instead of racing.
- **The restart is verified by pid, not by reachability.** `gpg-connect-agent /bye` succeeding only proves that *an* agent answers; when the wait cap was exhausted with the old agent still alive, it answered and the check passed while the new config was not applied. The hook now re-reads the pid and warns when it is unchanged or missing, and the warning names both halves of the recovery (`gpgconf --kill` **and** `gpg-connect-agent /bye`) rather than only the kill that already ran.
- **The sibling unbounded call is bounded.** `cli/zsh.nix` ran `gpg-connect-agent updatestartuptty /bye` with no timeout on the SSH interactive-shell path, where a wedged agent hangs every new remote shell — a wider blast radius than the hook. Wrapped in `timeout` by store path, since `timeout` is not otherwise guaranteed on PATH there.

Also took both style points: the `gawk` dependency is replaced by a `read` that asserts the `D` tag, and a comment explains that `kill` is left unqualified deliberately, being a builtin that needs no PATH.


yqrashawn:
## pr-review-loop — pass 1

`6f0437a51645` — **MERGEABLE**

Read the diff, the surrounding source, the live home-manager activation script and the nix-darwin-generated `set-environment` on disk.

**Verified by RUNNING:** `bash -n` on the extracted hook (syntax OK) and a stub-binary simulation of the hook under `set -eu -o pipefail` with four probe outputs — that is what confirms findings 1 and 3. Everything else is a reading of the source; finding 2's trigger is traced through `/nix/store/...-set-environment`, which `/etc/zshenv` sources.

VERDICT[23b5d5e4]: MERGEABLE — 6 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:150 — empty-pid branch trades the race for a fixed `sleep 1`, and the new pid verification structurally cannot catch it. TRACED + simulated: when `gpgAgentOld=""` and the restart yields any pid, `[ -z "$gpgAgentNew" ]` is false and `[ "$gpgAgentNew" = "" ]` is false, so no warning ever fires on this path. The PR body says this branch "no longer races"; it still does — 1s is an unmeasured guess at the same "stopped answering, sockets not yet unlinked" window the pid wait exists to avoid, and it is the one path where failure is silent. Trigger: probe fails while an agent is still bound (wedged, or mid-shutdown). Direction: derive the signal from the socket (`gpgconf --list-dirs agent-socket` plus an existence poll) or from `pgrep -u "$USER" gpg-agent`, so the branch has the same evidence the pid branch has.
2. [correctness/followup] modules/yqrashawn/darwin/core.nix:111 — the widest unbounded `updatestartuptty` sibling is untouched, and it runs *before* the one this PR bounded. `programs.gnupg.agent.enableSSHSupport = true` makes nix-darwin emit `gpg-connect-agent --quiet updatestartuptty /bye` into `environment.extraInit`; traced to `/nix/store/20646dmln1y1y76kzsh1makdrzcqrpv1-set-environment:21`, which `/etc/zshenv:11` sources for every shell — interactive or not, local or remote. Second sibling: `.doom.d/better-default.el:765` `(shell-command "gpg-connect-agent updatestartuptty /bye >/dev/null")`, synchronous, in `use-package! pinentry :config`, so a wedged agent hangs Emacs startup unbounded. Reported as one finding per the sibling rule; the zsh.nix one is the least reachable of the three.
3. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:136 — `read -r tag pid _` inspects only the first stdout line, where the replaced `awk '/^D /{print $2; exit}'` scanned for the first `D` line anywhere. Simulated: feeding `OK\nD 1234` yields an empty pid under the new code and `1234` under the old. Whether `gpg-connect-agent --no-autostart 'GETINFO pid' /bye` can ever emit a line before the `D` is INFERRED, not traced — I did not connect to the agent to check. A failed parse degrades to the finding-1 branch rather than erroring, so this is cheap to harden (`while read -r tag pid _; do [ "$tag" = D ] && { echo "$pid"; break; }; done`) and cheap to leave.
4. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:131 — "the bash builtin needs no PATH, which home-manager's activation script does not provide" is false. TRACED: `/nix/store/wpngh65ca25wlfvi954amp706yy4vgaq-activation-script:7` exports a PATH containing coreutils, gnugrep, gnused, findutils, jq and nix before any onChange hook runs. The code is fine — `kill` is a builtin — but the stated reason is wrong, and a future reader will use it to justify leaving other commands unqualified.
5. [docs-accuracy] modules/yqrashawn/home-manager/cli/zsh.nix:189 — "a wedged agent would otherwise hang every new remote shell" does not hold as a justification for this timeout. The identical call at `set-environment:21` already ran, unbounded, from `/etc/zshenv`, before `.zshrc` is read; a wedged agent hangs the shell there regardless. The call itself is still needed (the user's `~/.zshenv` sets `PINENTRY_USER_DATA` after `/etc/zshenv` runs) — it is the hang claim that is false. PR body's "The sibling unbounded call is bounded" overstates by the same margin; two siblings remain (finding 2).
6. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:126 — sixth PR on this hook, now 30 lines of probe/kill/wait/verify for a config reload, and the race being patched is created by the hook's own decision to restart immediately (`default.nix:119-122`). Worth asking whether `gpgconf --kill gpg-agent` followed by `gpgconf --launch gpg-agent` (idempotent, no overlap window) removes the need for the pid dance entirely (#subtract). The stated reason to restart eagerly — keeping `SSH_AUTH_SOCK` live in open shells — is a good one and the current design does serve it, so this is a question, not a defect.

PR #163, opened 2026-09-16
refactor(gpg): use gpgconf --launch, stop Emacs blocking on the agent

yqrashawn:
Follow-up to #162. Its style finding asked the right question — six PRs in, the hook had grown to 30 lines of probe/kill/wait/verify, and every correctness finding since #160 has been about a race that the hook's own hand-rolled restart created. So this takes the subtraction instead of patching the dance again.

**`gpgconf --launch gpg-agent` replaces the pid probe, the wait loop, the empty-pid fallback sleep and the pid comparison.** It is idempotent and does its own waiting, so there is no window where a replacement binds sockets that the still-dying agent then unlinks — which removes findings 1 and 3 of #162 rather than hardening them. Both calls stay bounded by `timeout` and guarded, and a failure still prints a warning naming the recovery command. Net 20 lines removed.

**The remaining unbounded sibling that is ours is now async.** `.doom.d/better-default.el` called `gpg-connect-agent updatestartuptty` through synchronous `shell-command` in `use-package! pinentry :config`, so a wedged agent hung Emacs startup. It is now `start-process-shell-command`; nothing there uses the result. The third sibling, `environment.extraInit` in nix-darwin's `programs.gnupg.agent`, is generated upstream and not editable from this repo — noted, not fixed.

Also corrected the two comments #162 flagged as false: the activation script does provide a PATH (`kill` is still a builtin, but that was the wrong reason), and the zsh-side timeout bounds only the repeat we add, since `/etc/zshenv` already ran the same call unbounded.


yqrashawn:
## pr-review-loop — pass 1

`e110a0d5f694` — **MERGEABLE**

Read the diff, the surrounding source, the base version of the hook, and the nix-darwin module that generates the third call site.

VERDICT[8ebb0f3f]: MERGEABLE — 5 follow-ups to file

      [correctness/blocking]  none
      [coverage]              none
      [style]                 none

1. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:132 — `gpgconf --launch` exits 0 whenever it can connect to *any* agent, including the old one that is still shutting down after line 131's `--kill`. `gpgconf --kill` sends KILLAGENT and returns without waiting for the process to exit (the premise of #160/#161, unchanged here), and the dying agent keeps its listener bound until `agent_exit`. If `--launch` connects in that window it starts nothing, exits 0, the warning at :133 cannot fire, and after the old agent exits **no agent is running at all, silently**. The base's pid comparison (54ad4b04d, default.nix:152-157) detected exactly this case and is deleted here with no replacement. Trigger: `gpg-agent.conf` content changes *and* the old agent outlives three `fork`/`exec`s (`timeout`+`gpgconf`+`gpg-connect-agent`) — a narrow window, hence followup, not blocking. Traced from `man gpgconf` ("If the component is not already running, start it") plus the `NOP` / `--no-autostart` / `KILLAGENT` strings in the gnupg 2.4.8 `gpgconf` binary, which show both subcommands are thin `gpg-connect-agent` wrappers. Recovery is a new shell or any gpg command, so impact is bounded.

2. [correctness/followup] .doom.d/better-default.el:768 — the only one of the three call sites with no bound at all: zsh.nix:193 has `timeout 5`, default.nix:131-132 have `timeout 10`, this has none. Against a wedged agent it now leaks one `gpg-connect-agent` process per Emacs start instead of hanging init. Not a hang: `confirm-kill-processes` is nil at better-default.el:17, so the stray process does not block `kill-emacs` either. Low impact, but the fix is the same `timeout` the two siblings already use.

3. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:126 — "it is idempotent and does the waiting itself, which removes the window where a replacement binds sockets that the still-dying agent then unlinks" is false. `--launch` has no knowledge of the agent `--kill` just killed and performs no wait; it is `gpg-connect-agent NOP`, which autostarts only when `connect()` fails. Whatever narrowing of the window exists comes from that connect-then-autostart ordering, not from waiting — a different mechanism, and it is what leaves finding 1 undetected. This matters because the false claim is the stated justification for deleting the verification.

4. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:122 — "the ssh socket that SSH_AUTH_SOCK points at in already-open shells comes back immediately". The comment block is rewritten in this diff and the PR body reasons from this claim, so it is in scope. The gpg-agent.conf generated four lines above (:111-117) has no `enable-ssh-support`, and nix-darwin's gnupg module writes none either — so no agent this hook starts ever creates `S.gpg-agent.ssh`. Independently, prezto's `ssh` pmodule (modules/yqrashawn/home-manager/cli/prezto.nix:25) points `SSH_AUTH_SOCK` at `~/.cache/prezto/ssh-agent.sock` in interactive shells, after `/etc/zshenv` set it to the gpg path. Verified by running `ls ~/.gnupg/S.*` (no sockets exist) and `echo $SSH_AUTH_SOCK` (prezto's path). If that holds, nothing depends on the agent coming back before the next autostart, and `gpgconf --kill` alone is the whole hook — which subtracts findings 1 and 3 rather than patching them.

5. [docs-accuracy] modules/yqrashawn/darwin/core.nix:111 — the PR body says the third sibling, "`environment.extraInit` in nix-darwin's `programs.gnupg.agent`, is generated upstream and not editable from this repo". It is generated by `enableSSHSupport = true`, set on this line by this repo: nix-darwin's `modules/programs/gnupg.nix` emits the unbounded `gpg-connect-agent --quiet updatestartuptty /bye` only inside `optionalString cfg.agent.enableSSHSupport`. Confirmed present on disk at `/nix/store/…-set-environment`, sourced by `/etc/zshenv`. So the remaining unbounded call — the one that hangs *every* new shell against a wedged agent, the worst of the three — is reachable from this repo by flipping that option, which per finding 4 appears to buy nothing today.

Method: findings 1 and 3 rest on `man gpgconf` and `strings` over the installed gnupg 2.4.8 `gpgconf` binary (run), plus reading the diff and base. Findings 4 and 5 rest on reading nix-darwin's gnupg module and the generated `set-environment`/`/etc/zshenv` (read), confirmed by running `ls ~/.gnupg/S.*`, `ps`, and `echo $SSH_AUTH_SOCK` on this host — one host's live state, so treat the prezto override as evidence, not proof. Finding 2 is a reading of the source. I did not start an agent or run `gpgconf --launch`; gnupg's internal exit/cleanup ordering is inferred from the documented semantics above, not traced in gnupg source, which is not present in this worktree. No test exists for either file, so there is no coverage finding to make.

PR #164, opened 2026-09-16
refactor(gpg): kill the agent and let autostart handle the rest

yqrashawn:
Follow-up to #163, taking its finding 4 — the one that dissolves the rest.

Every race reported since #160 came from the hook restarting the agent eagerly. The stated reason was keeping `SSH_AUTH_SOCK` live in already-open shells. That reason does not hold here: the generated `gpg-agent.conf` sets no `enable-ssh-support`, and `SSH_AUTH_SOCK` is `~/.cache/prezto/ssh-agent.sock` (prezto's `ssh` pmodule, `cli/prezto.nix`), not a gpg-agent socket — checked on this host.

So the hook is now `gpgconf --kill gpg-agent` and nothing else. The next gpg command autostarts the agent with the new config. That removes, rather than patches, both correctness findings on #163: there is no `--launch` that can silently attach to the dying agent and start nothing, and no restart to verify. Still bounded by `timeout` and still guarded, with a warning if the kill itself fails.

Dropping #163's remaining points: the `--launch` comment it called false is deleted along with the call, and the Emacs `updatestartuptty` is already async as of #163, so a wedged agent leaks one short-lived process rather than hanging init — not worth a `timeout` whose binary is not guaranteed on PATH there. Flipping `enableSSHSupport` off in `darwin/core.nix` to remove the last unbounded call site looks right for the same reason this PR works, but it changes what the agent serves and belongs in its own change with the user's say-so.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GPG agent configuration handling to stop the agent safely without automatically relaunching it.
  * Added a timeout and warning when the agent cannot be stopped, indicating that updated settings may apply after the agent exits.
  * GPG commands now start the agent automatically when needed.


yqrashawn:
## pr-review-loop — pass 1

`790e54e7cd60` — **MERGEABLE**

Read the diff, the surrounding nix, `darwin/core.nix`, the live nix-darwin `set-environment`, and prezto's `ssh` module from the store. I **ran** exactly one thing: `gpgconf --kill gpg-agent` against a throwaway `GNUPGHOME` (no agent there) — it exits 0 silently, so the `|| echo` guard does not fire on "no agent running", only on a hang (`timeout` → 124) or exec failure. Everything else below is a reading of the source; the one inference is labelled.

VERDICT[de04fecb]: MERGEABLE — 3 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none

1. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:124 — "SSH_AUTH_SOCK is pointed at prezto's own ssh-agent socket (cli/prezto.nix), not at gpg-agent's" is false as the system sets it, and the PR body repeats it. TRACED: `modules/yqrashawn/darwin/core.nix:111` sets `programs.gnupg.agent.enableSSHSupport = true`, and the live nix-darwin env script (`/nix/store/20646dmln1y1y76kzsh1makdrzcqrpv1-set-environment:21-23`, sourced by `/etc/zshenv` for *every* shell) does `gpg-connect-agent --quiet updatestartuptty /bye` and `export SSH_AUTH_SOCK=$(gpgconf --list-dirs agent-ssh-socket)` — i.e. SSH_AUTH_SOCK *is* gpg-agent's ssh socket path. Prezto only overrides it in interactive zsh, and only as a fallback: its `ssh` module guards on `[[ ! -S "$SSH_AUTH_SOCK" ]]`, which is true here solely because this conf omits `enable-ssh-support` so that socket never exists. The conclusion the hook rests on survives, but the stated reason is the wrong one, and it is load-bearing in the file it sits next to: add `enable-ssh-support` to this same conf later and the claim silently inverts — prezto would then stop overriding, SSH_AUTH_SOCK would resolve to gpg-agent, and kill-without-restart would break ssh in already-open shells, the exact hazard the deleted comment named. Also worth noting for the same reason: on this host it is not "the next gpg command" that brings the agent back, it is the next shell.

2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:129 — kill-only drops the wait added in 1531bddf3 and the verify added in 54ad4b04d, leaving a residual (narrower) window: `gpgconf --kill` returns when the KILLAGENT connection closes, before the dying agent runs its socket cleanup. TRACED: the only autostart call sites are the per-shell `gpg-connect-agent updatestartuptty /bye` in set-environment:21 and `modules/yqrashawn/home-manager/cli/zsh.nix:190`; either can fire during that window if a shell starts while activation runs, and nothing in the hook waits for the old process. INFERRED (not verified against gnupg source here): gpg-agent's cleanup unlinks its socket path unconditionally, so a replacement that bound the socket in that window has it unlinked under it, leaving the socket absent until the next autostart. Sub-millisecond trigger, self-healing on the following gpg call, and strictly narrower than the `--launch` race it replaces — followup, not blocking. Same root: exit 0 from `gpgconf --kill` is not evidence the agent exited, so the hook can report success with the old config still live.

3. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:130 — the warning drops both the diagnosis and the remedy. `>/dev/null 2>&1` at :129 discards gpgconf's own error, so a user who sees this line cannot tell a `timeout` hang from an exec failure, and unlike the message it replaces ("run 'gpgconf --launch gpg-agent'") it names no action. Sending gpgconf's stderr through, or distinguishing exit 124, would make the one line the hook can print worth printing.

PR #165, opened 2026-09-16
docs(gpg): state why kill-only is safe, make the warning actionable

yqrashawn:
Follow-up to #164. Comment and warning only — no behavior change.

- **The load-bearing comment gave the wrong reason.** It said `SSH_AUTH_SOCK` is prezto's socket rather than gpg-agent's. As the system sets it, nix-darwin's `extraInit` exports `SSH_AUTH_SOCK=$(gpgconf --list-dirs agent-ssh-socket)` for every shell, because `darwin/core.nix` sets `enableSSHSupport`; prezto only wins because its `ssh` module falls back when that socket does not exist, which is true solely because this conf omits `enable-ssh-support`. The conclusion holds, but the condition it depends on now sits in the comment, next to the option that would invert it.
- **The residual race is written down instead of engineered around.** `gpgconf --kill` returns on acknowledgement, not on exit, so a shell starting in that window can autostart a replacement whose socket the dying agent unlinks. Sub-millisecond and self-healing on the next gpg call; #160 through #163 are the record of what chasing it costs.
- **The warning says something useful.** gpgconf's own stderr is no longer discarded, and the message reports the exit code (calling out 124 as the hang) and names the retry.

Also corrected: the agent comes back on the next shell here, not only on the next gpg command.


yqrashawn:
## pr-review-loop — pass 1

`81eb2b2711f8` — **MERGEABLE**

Verified by RUNNING: `nix-instantiate --parse` on the changed file (parses), and `gpgconf --kill gpg-agent` against an empty temporary `GNUPGHOME` (exit 0, silent — no agent started, user's live agent untouched). Everything else is a reading: `/nix/store/...-set-environment` (nix-darwin `extraInit`), prezto's `modules/ssh/init.zsh`, the installed home-manager `activate` script (`set -eu -o pipefail`, onChange inlined verbatim inside `if [[ ! -v DRY_RUN ]]`).

Traced and confirmed correct in the new comment: `extraInit` does run `gpg-connect-agent --quiet updatestartuptty /bye` and does `export SSH_AUTH_SOCK=$(gpgconf --list-dirs agent-ssh-socket)` because `modules/yqrashawn/darwin/core.nix:111` sets `enableSSHSupport`; prezto's `ssh` module takes over only via `if [[ ! -S "$SSH_AUTH_SOCK" ]]`, which holds solely because the conf omits `enable-ssh-support`. `$?` in the `||` branch is the `timeout` status (expansion precedes the `>&2` redirection). No launchd gpg-agent job exists, and `gpgconf` is on PATH at `/run/current-system/sw/bin`, so the retry hint is runnable.

VERDICT[a3815236]: MERGEABLE — 6 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none

1. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:137 — PR body opens "Comment and warning only — no behavior change", but `2>&1` was dropped: gpgconf's and timeout's stderr now reach activation output. The body's own third bullet says so, contradicting its first line. Traced in the diff.
2. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:134 — "Self-healing on the next gpg call" understates the residual. If the replacement agent's socket is unlinked by the dying one, that replacement keeps running on an unreachable inode; the next gpg call autostarts a *third* agent and the second lingers until logout. Function heals, process leaks. INFERRED from gnupg's stale-socket handling, not traced in source (gnupg source not in this worktree).
3. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:138 — "(exit $?; 124 means it hung)" implies a readable range of codes, but `gpgconf --kill` exits 0 even when the kill does not take (VERIFIED: exit 0 against an empty GNUPGHOME; `gc_error` is called with status 0 in `agent_runtime_change`, so it warns without exiting). Only 124 (timeout) and 126/127 (exec failure) can reach the branch.
4. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:133 — "sub-millisecond" is an unmeasured quantity in a load-bearing comment; nothing in the diff or history instantiates it. "Brief" or a stated bound would carry the same weight honestly.
5. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:123 — "the next shell" is imprecise: `/etc/zshenv` guards `set-environment` on `__NIX_DARWIN_SET_ENVIRONMENT_DONE`, which `set-environment` exports, so child shells skip it. Only a fresh top-level shell (new terminal) re-runs `updatestartuptty`.
6. [correctness/followup] .doom.d/better-default.el:768 — sibling of the bounding pattern this comment is about: `start-process-shell-command "gpg-updatestartuptty" ... "gpg-connect-agent updatestartuptty /bye"` has no `timeout`, unlike `cli/zsh.nix:191` (`timeout 5`) and this hook (`timeout 10`). Trigger: wedged agent — Emacs does not block (async), but the `gpg-connect-agent` process never exits. Pre-existing, outside this diff.

PR #166, opened 2026-09-16
fix(gpg): bound the Emacs updatestartuptty process

yqrashawn:
Follow-up to #165, its one correctness finding — the last unbounded `gpg-connect-agent` call that this repo owns.

`cli/zsh.nix` wraps its call in `timeout 5` and the activation hook in `timeout 10`; the Emacs one had no bound. Since #163 it is async, so a wedged agent no longer hangs init, but the `gpg-connect-agent` process itself never exits and sticks around for the life of the Emacs session.

Bounded with a 5-second timer that kills the process if it is still live, matching the shell call sites. An elisp timer rather than a `timeout` wrapper because `timeout` is not guaranteed on Emacs's `exec-path` the way it is in the Nix-built call sites.

The remaining unbounded call is nix-darwin's own `extraInit`, which is generated by `enableSSHSupport` and not editable in place; #164 noted that turning that option off is the real fix and wants the user's say-so.

Dropping #165's documentation and style points — they are about the previous PR's own comment wording, and it now says what the code does.


yqrashawn:
## pr-review-loop — pass 1

`5a47c0aeb852` — **MERGEABLE**

VERDICT[c8af9a8d]: MERGEABLE — 2 follow-ups to file

      [correctness/blocking]  none
      [correctness/followup]  none
      [coverage]              none
      [docs-accuracy]         1
      [style]                 1

1. [docs-accuracy] .doom.d/better-default.el:768 — "without it a wedged agent leaves the process around until logout" is wrong about the bound. Emacs signals its own subprocesses at exit (`kill_buffer_processes` on shutdown, plus pty hangup), so the unbounded case lasted the Emacs session, not until logout. The PR body states this correctly ("for the life of the Emacs session"); only the in-file comment overstates it.
2. [style] .doom.d/better-default.el:769-771 — the shell wrapper is now dead weight and it is what makes the kill non-obvious. With `BUFFER` nil, Emacs already discards stdout/stderr, so `>/dev/null 2>&1` buys nothing but forces `start-process-shell-command`, which spawns `bash -c` with `gpg-connect-agent` as a *grandchild* (traced: `ps` shows two PIDs, no exec optimisation). `(start-process "gpg-updatestartuptty" nil "gpg-connect-agent" "updatestartuptty" "/bye")` kills exactly the process it names.

Verification method — what I RAN vs READ:

RAN (`emacs -Q --batch`, no servers, no services, throwaway worktree):
- Killing the shell process does reach the `gpg-connect-agent`-shaped grandchild. Stand-in `sleep`: before kill, `bash -c` at pid P (pgid P) plus child at pgid P; after `kill-process`, both gone. Holds for `bash -c` and `/bin/sh -c`, for a simple command with redirections and for two commands, and with `process-connection-type` nil as well as the pty default. So the PR's central claim — a wedged client no longer lingers — is delivered, not just asserted.
- Emacs's own pgid stays distinct from the child's, so the group signal cannot reach Emacs itself.
- `kill-process` on an already-exited process errors (`"Process d is not active"`), which is why the `process-live-p` guard matters; it is present at :774. There is no wait/redisplay point between the guard and the kill inside that lambda, so the status cannot flip between them in single-threaded Emacs — no reachable race, no finding.
- The timer/closure path as written: fires, kills, leaves `process-live-p` nil.

READ:
- `.doom.d/better-default.el:1` is `-*- lexical-binding: t; -*-`, so the `lambda` at :773 genuinely closes over `proc`; under dynamic binding this would have been a void-variable at timer fire. Worth stating because the whole fix hangs on it.
- Sibling sweep, repo-wide (`gpg-connect-agent|gpgconf`, all files): `cli/zsh.nix:191` `timeout 5`, `dotfiles/default.nix:137` `timeout 10`, `better-default.el:776` commented out. No other unbounded call site in-tree, so "the last unbounded call that this repo owns" checks out; nix-darwin's generated `extraInit` is out of tree, as the body says.
- 5s matches `timeout 5` in `zsh.nix:191`. `kill-process` is SIGKILL where `timeout` defaults to SIGTERM — for a client whose only state is an Assuan round trip this is not a defect, so no finding.
- No `better-default-test.el` exists anywhere in `.doom.d/`, and none of this file's 1270 lines is covered. That is a pre-existing gap in a startup-side-effect file rather than a test that certifies something it does not check, so it is not a [coverage] finding under the definition here.

One pre-existing question I deliberately did not file, since I could not verify it and it predates this diff: `GPG_TTY` is exported only from zsh's interactive `initContent` (`cli/zsh.nix:178`), not from `.zshenv`, so a GUI/launchd Emacs runs `updatestartuptty` with no `GPG_TTY` and the agent would record whatever tty the subprocess happens to have. Reproducing that needs a GUI Emacs under launchd, which I don't have here — flagging it as a hypothesis only, not a finding against this change.

PR #167, opened 2026-09-16
chore(gpg): raise the ssh cache TTL to 60 days

yqrashawn:
Follows #156, which raised the gpg passphrase cache to 60 days. The ssh-side options were unset and so fell back to gpg-agent's defaults of 1800s idle / 7200s absolute. Both are now 5184000 seconds, matching the gpg side.

Verified the file parses as a config: `gpg-agent --homedir <scratch> --gpgconf-test` exits 0 with these options and the comment block.

One thing the change does not do, deliberately. These options only bite when gpg-agent serves the ssh-agent protocol, and this conf omits `enable-ssh-support`, so it currently serves no ssh socket and `SSH_AUTH_SOCK` resolves to prezto's own ssh-agent. The settings are therefore inert until `enable-ssh-support` is added here — which is a behavior change (it moves ssh key handling from prezto's agent, seeded from the macOS keychain via `ssh-add --apple-load-keychain`, onto gpg-agent) and wants a separate decision, so it is not bundled in. The comment in the file says so at the point where someone would flip it.


yqrashawn:
## pr-review-loop — pass 1

`42d4f46b31a0` — **MERGEABLE**

Read the diff, the surrounding file, nix-darwin's gnupg module, prezto's ssh module, the generated `/etc/zshenv` + `set-environment`, and the gpg-agent man page.

**Verified by RUNNING:** config parse (`gpg-agent --homedir <scratch> --gpgconf-test` → exit 0, author's claim reproduced), `gpg-agent --dump-options | grep cache-ttl` (all four options exist), `awk` over `~/.gnupg/sshcontrol` (5 entries, TTL field `0`, no flags), grep of `/etc/zshenv` and `~/.zshrc` ordering. **Everything else is a reading**: traced `core.nix:111` → nix-darwin `modules/programs/gnupg.nix` → `set-environment:23` → prezto `modules/ssh/init.zsh:23`. No inferences-from-name in the list below.

Body claims that check out: ssh defaults were 1800s/7200s (man page lines 288, 302); `enable-ssh-support` absent so no ssh socket served; prezto loads at `~/.zshrc:31` *before* the keychain load at `:86`, so interactive login shells behave as described. Also checked the sshcontrol per-key TTL override (man: "A non-zero TTL overrides the global default") — all five entries are `0`, so the new default would govern. Not a finding.

VERDICT[f15d3e93]: MERGEABLE — 3 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  1
  [coverage]              none
  [docs-accuracy]         1
  [style]                 1

1. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:116 — "SSH_AUTH_SOCK falls through to prezto's own ssh-agent" is false outside interactive zsh: `/etc/zshenv` sources `set-environment:23`, which runs `export SSH_AUTH_SOCK=$(gpgconf --list-dirs agent-ssh-socket)` unconditionally (nix-darwin gnupg.nix gates only on `enableSSHSupport`, set at darwin/core.nix:111 — not on this conf). Every shell therefore gets `~/.gnupg/S.gpg-agent.ssh`, a path that does not exist (no socket in `~/.gnupg`); only prezto's `[[ ! -S "$SSH_AUTH_SOCK" ]]` check (zsh-prezto `modules/ssh/init.zsh:23`, interactive zsh only) repoints it. Non-interactive zsh, bash scripts and launchd/GUI-spawned processes keep the dead path — it is an override by prezto, not a fall-through, and it does not happen there.
2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:118 — trigger: someone adds `enable-ssh-support` as the comment invites. `max-cache-ttl-ssh 5184000` then makes each of the 5 sshcontrol keys usable by any local process with no prompt for 60 days, since no entry carries the `confirm` flag (verified: `fields=2`, no flags column) and nothing in the repo sets one. The PR pre-wires the TTL half of the deferred decision but not its guard, so the flip is a larger posture change than the comment's "behavior change … wants a separate decision" conveys (#negative-space). Inert today.
3. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:112 — `5184000` is now a bare literal in four places. The drift this PR fixes (#156 moved two, the ssh pair stayed at gpg's defaults) is caused by that duplication; a `let ttl = "5184000";` above `text` removes the class rather than the instance (#subtract).

One trade worth naming, not a finding: touching this file fires the `onChange` kill at default.nix:139, so applying the PR drops the currently-cached gpg passphrase to install settings that do nothing until `enable-ssh-support` lands.

PR #168, opened 2026-09-16
refactor(gpg): bind the cache TTL once, name the ssh-support prerequisite

yqrashawn:
Follow-up to #167.

- **One `cacheTtl` binding instead of four literals.** The drift #167 fixed — the ssh pair sitting at gpg-agent's 1800s/7200s defaults while the gpg pair was already at 60 days — existed because the value was duplicated and one copy got edited. Binding it once removes the class rather than the instance.
- **The comment now names the guard, not just the flip.** #167 invited a later `enable-ssh-support` without saying what else that needs: no entry in `~/.gnupg/sshcontrol` carries the `confirm` flag, so with a 60-day `max-cache-ttl-ssh` every listed key would be usable by any local process with no prompt for the full period. The comment says to add `confirm` in the same change.
- **Corrected the SSH_AUTH_SOCK description in both comments.** It is not a fall-through: nix-darwin's `extraInit` exports the gpg-agent ssh socket path in every shell because `darwin/core.nix` sets `enableSSHSupport`, and prezto then overrides it — only in interactive zsh. Non-interactive shells and launchd-spawned processes keep the dead path.

No behavior change: the rendered `gpg-agent.conf` options are the same values as on master.


yqrashawn:
## pr-review-loop — pass 1

`3936252834bc` — **MERGEABLE**

Verification method: findings 1, 2 and the *absence* of a behavior finding were verified by RUNNING things (`git show 188ec4cc7`, `git log -S cache-ttl`, `nix eval` on both rendered text blocks, `nix-instantiate --parse`, `nixpkgs-fmt` drift comparison base-vs-head, `grep` of `/etc/zshenv` → `/nix/store/…-set-environment:23`, the gnupg 2.4.8 info SIGHUP option list, and a count-only `grep` of `~/.gnupg/sshcontrol`). Findings 3–6 are a reading of the source.

Confirmed no behavior change, as claimed: rendered option lines are byte-identical to base (`diff` of the two `nix eval --raw` outputs, comments excluded), the file parses, and formatter drift is 11 lines on both base and HEAD — this PR adds none. The `allow-loopback-pinentry`-is-not-SIGHUP-rereadable claim at :141 checks out against gnupg 2.4.8's own SIGHUP list, and the `SSH_AUTH_SOCK`-in-every-shell claim at :124 is exactly right: `set-environment:23` exports it unconditionally and `/etc/zshenv` sources that for every `rcs` zsh.

VERDICT[af39935f]: MERGEABLE — 6 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  none
  [coverage]              none
  [style]                 4 findings

1. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:112 — causal claim false; drift was omission, not stale duplicate. TRACED: `git show 188ec4cc7` (#156) is `+2 -2`, editing *both* gpg literals together (604800→5184000); `git log -S"cache-ttl"` shows `default-cache-ttl-ssh`/`max-cache-ttl-ssh` first appearing in 5a0fa1aa5 (#167). No ssh literal ever existed to be "edited on its own". Same wrong mechanism in the PR body ("existed because the value was duplicated and one copy got edited... removes the class rather than the instance") — a single binding does not prevent the class that actually bit, which is forgetting to emit an option at all.
2. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:123 — PR body's extension of this comment ("launchd-spawned processes keep the dead path") is false. TRACED: `/etc/zshenv` sources `set-environment` only under `[[ -o rcs ]]`, and `grep -rn "launchctl setenv"` finds SSH_AUTH_SOCK set nowhere (only PATH/proxy/token vars in `darwin/openclaw.nix:79-92`). Launchd-spawned processes have SSH_AUTH_SOCK **unset**, not pointed at a dead socket. The in-file comment itself is correct; only the body overreaches.
3. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:113 — `cacheTtl = "5184000"` is still a magic literal whose meaning lives only in adjacent prose. The PR's own goal is to stop a value and its intent diverging; `toString (60 * 24 * 3600)` makes the binding self-checking and removes the last pair that can disagree. (#subtract)
4. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:128 — unnamed coupling: one binding forces gpg and ssh TTLs equal forever, while this very comment argues the ssh cache carries strictly more exposure and needs a `confirm` guard the gpg path has no equivalent for. gpg-agent's own defaults differ (600/7200 vs 1800/7200) for that reason. If the ssh TTL is later lowered on the comment's own logic, the binding must be split again. (#decompose)
5. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:131 — "Add the confirm flag to sshcontrol in the same change" prescribes a guard in state this repo does not manage: `grep -rn sshcontrol --include=*.nix` matches only these comments, and there is no `home.file` entry for `.gnupg/sshcontrol`. So it cannot be part of "the same change" in the Nix sense — it is a manual edit `darwin-rebuild` neither applies nor verifies. The factual premise does hold (RAN a count-only grep: 5 non-comment entries, 0 with `confirm`), but it describes unmanaged user-local state and will rot silently. Managing `.gnupg/sshcontrol` declaratively would make the prerequisite enforceable rather than advisory.
6. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:119 — the security note is written entirely for the inert ssh path and silent about the live gpg pair with the identical property: `default-cache-ttl`/`max-cache-ttl` at 60 days mean any local process can drive the agent socket to sign or decrypt with no prompt today, and there is no per-key `confirm` equivalent for the gpg protocol. Reads as if the exposure is entirely hypothetical. Also a second, more durable reason for kill-not-reload that :141 omits: `default-cache-ttl-ssh`/`max-cache-ttl-ssh` are absent from gnupg 2.4.8's SIGHUP-honored option list, so they would survive a `--reload` even if `allow-loopback-pinentry` were someday dropped. (#negative-space)

PR #169, opened 2026-09-17
docs(gpg): the agent already serves ssh, so the ssh TTLs are live

yqrashawn:
Comment-only correction. #167 and #168 both asserted that gpg-agent serves no ssh socket because this conf omits `enable-ssh-support`, and called the ssh TTLs inert. That is wrong on this machine, and I checked it directly:

- `lsof -U` shows the running gpg-agent (pid 4286, started after the last rebuild, reading the current conf) owning `~/.gnupg/S.gpg-agent.ssh`.
- `SSH_AUTH_SOCK=~/.gnupg/S.gpg-agent.ssh ssh-add -l` lists keys and exits 0.
- `~/.cache/prezto/ssh-agent.sock` — the path prezto exports in interactive shells — is a hand-made symlink to that same gpg-agent socket.
- `~/.gnupg/sshcontrol` has 31 lines, none with the `confirm` flag.

So the 60-day `*-cache-ttl-ssh` values added in #167 are in force right now, not waiting on a future switch, and the security note that was written as a warning about a hypothetical flip describes the current state.

The comments now say that, and the `onChange` comment says what the kill actually costs: it drops the ssh key cache as well as the gpg one and removes the socket `SSH_AUTH_SOCK` names until the agent is autostarted again.

Also corrected the `cacheTtl` rationale: the ssh pair was never a stale duplicate, it was simply absent until #167.

## Summary by CodeRabbit

* **Documentation**
  * Clarified comments describing GPG agent cache duration and SSH key handling.
  * Documented the effects of reloading the GPG agent, including SSH cache and socket cleanup.
  * No user-facing configuration or behavior changes.


yqrashawn:
## pr-review-loop — pass 1

`1b7008e06dbb` — **NOT MERGEABLE**

Verified by reading the diff and source, plus **read-only** shell inspection of files already on disk (`~/.gnupg/gpg-agent.conf`, `~/.gnupg/sshcontrol` field structure only, `~/.cache/prezto/ssh-agent.sock`, `/nix/store/…-set-environment`, the active prezto `modules/ssh/init.zsh`, and the GnuPG 2.4.8 info manual in the store). I did **not** start, connect to, or query gpg-agent — no agent was running during the review — and I ran no test suite.

The PR's central correction is **right**, and the manual backs it more strongly than the PR's own evidence does: `gnupg.info-1:600` — "On Unix platforms the OpenSSH Agent protocol is always enabled, but `gpg-agent` will only set the `SSH_AUTH_SOCK` variable if the option `enable-ssh-support` is given." So #167/#168's "inert" premise was false, and this PR is correcting it in the right direction.

VERDICT[109a6536]: NOT MERGEABLE — the 60-day ssh TTLs were justified solely by "inert until enable-ssh-support"; this PR retracts that premise, confirms 5 passphrase-free ssh keys usable by any local process for 60 days, and leaves the values unchanged.

      [correctness/blocking]  1 findings
      [coverage]              none

1. [correctness/blocking] modules/yqrashawn/home-manager/dotfiles/default.nix:132 — The only stated rationale for `default-cache-ttl-ssh`/`max-cache-ttl-ssh` at 5184000 was #167's "inert until enable-ssh-support"; this PR retracts that premise and keeps the value. Traced, not inferred: `~/.gnupg/sshcontrol` is 31 lines = 25 comment + 1 blank + **5** keygrip entries, every entry having exactly 2 fields with TTL `0`, and per `gnupg.info-1:825` only a *non-zero* per-entry TTL overrides `--default-cache-ttl-ssh` — so all 5 keys inherit 60 days, versus gnupg's 1800s/7200s defaults (`gnupg.info-1:428,440`), a 2880x widening of the no-prompt window. 0 entries carry `confirm` (`grep -c confirm` = 0). Consequence: after one unlock, any local process running as this user can authenticate/sign with those 5 keys, unprompted, for 60 days. The fix is in-repo and one line — these two lines are repo-managed even though `sshcontrol` is not. Sibling, reported here at the same severity but *not* requiring change: `:119-120` gives the same window to gpg operations, but that was a deliberate, separately-argued choice since #156; the ssh pair never was. Either lower the ssh pair to gnupg's defaults, or state an affirmative justification for 60 days on ssh in its own right.
2. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:149 — "the window is a single autostart" is contradicted by the prezto path it depends on. Traced through the *active* prezto (`/nix/store/3api06…-zsh-prezto-0-unstable-2025-07-30/share/zsh-prezto/modules/ssh/init.zsh`): an interactive zsh started after the kill with `__NIX_DARWIN_SET_ENVIRONMENT_DONE` already exported skips `set-environment`, so it keeps `SSH_AUTH_SOCK=~/.cache/prezto/ssh-agent.sock`, which now dangles; `init.zsh:23` `[[ ! -S "$SSH_AUTH_SOCK" ]]` is therefore true, `init.zsh:30` starts a plain `ssh-agent`, and `init.zsh:37` `ln -sf` **repoints the symlink away from gpg-agent**. That state persists for later shells — not self-healing on the next autostart. Trigger: a nested or non-login interactive zsh starting between the kill and the next agent autostart; narrow, hence followup rather than blocking, and pre-existing rather than introduced here.
3. [correctness/followup] modules/yqrashawn/home-manager/dotfiles/default.nix:121 — The prose is inside the `text = ''…''` heredoc, so it is part of the content-hashed `writeTextFile` output. Verified by reading the deployed `~/.gnupg/gpg-agent.conf`, which contains the *pre-PR* comment text verbatim. Editing comments therefore changes the store path, which fires `onChange` at `:156` and kills the agent — this documentation-only PR will itself flush both 60-day caches and unlink the ssh socket, the exact cost the diff spends thirty lines describing. Neither the comment nor the PR body mentions it. Moving the prose to Nix-level `#` comments outside the string would decouple it.
4. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:125 — "`~/.cache/prezto/ssh-agent.sock` is a hand-made symlink" is false. It is created automatically by prezto's ssh module, `init.zsh:37` `ln -sf "$SSH_AUTH_SOCK" "$_ssh_agent_sock"`, on every interactive shell where `SSH_AUTH_SOCK` is a live socket at a different path; the module is enabled at `modules/yqrashawn/home-manager/cli/prezto.nix:26`. The conclusion drawn from it ("interactive shells reach gpg-agent too") is correct; only the attribution is wrong, and it matters because a reader told the link is hand-made will not expect prezto to rewrite it (see finding 2).
5. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:150 — "every shell's rc runs `gpg-connect-agent` anyway" is false. Traced: the call lives in nix-darwin's `set-environment` (`/nix/store/20646…-set-environment:21`), which `/etc/zshenv` sources only under `if [ -z "${__NIX_DARWIN_SET_ENVIRONMENT_DONE-}" ]`, and that variable is `export`ed — so child and nested shells skip it. The only other call site, `modules/yqrashawn/home-manager/cli/zsh.nix:191`, is gated on `[[ -n "$SSH_CONNECTION" ]]`. Accurate wording would be "every new top-level login shell".
6. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:128 — The PR body's supporting claim, "`~/.gnupg/sshcontrol` has 31 lines, none with the `confirm` flag", uses a `wc -l` count as a measure of exposure. 31 lines is 25 comment lines, 1 blank, and 5 actual entries. The in-file comment at this line says "every key listed", which is accurate; the body reads as 31 exposed keys when the number is 5.
7. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:124 — "(verified with lsof on the running agent and `ssh-add -l` over that socket)" grounds a permanent invariant in a one-time, pid-specific observation. GnuPG documents it: `gnupg.info-1:600`. Citing the manual makes the comment survive the next agent restart, and would have prevented #167/#168 asserting the opposite.
8. [style] modules/yqrashawn/home-manager/dotfiles/default.nix:130 — #165 was specifically titled "make the warning actionable" and ended with the imperative "Add the confirm flag to sshcontrol in the same change." This PR replaces it with the passive "adding confirm there is a manual edit" at precisely the moment the risk stops being hypothetical. Keep an imperative.

yqrashawn:
Pushed a fix for the blocking finding and the rest.

**On the blocking finding.** The 60-day ssh TTL is not an orphaned value looking for a rationale — the owner asked for it directly ("also update the ssh one to 60 days"), and then, after being shown that the setting is live rather than inert and what it exposes, did not retract it. So the value stays and the file now carries an affirmative justification naming the owner's decision and the exposure it accepts, rather than the "inert" premise this PR retracted. The imperative #165 asked for is back: add `confirm` to the sshcontrol entries to get a prompt per use.

Verified the entry count myself rather than taking `wc -l` as exposure: 5 entries, all per-entry TTL 0, 0 with `confirm` — matching the review. Both the comment and this claim now say 5 entries, not 31 lines.

**Findings 2 through 8.** All taken:
- The prose moved out of the `text` heredoc into Nix-level comments, so a docs-only edit no longer changes the store path and no longer fires the kill (finding 3). That also makes this PR's own remaining conf diff empty of comment churn.
- The symlink is prezto's doing (`modules/ssh/init.zsh`), not hand-made (finding 4).
- The `onChange` comment now says what recovery actually looks like: only a new top-level login shell re-runs nix-darwin's `gpg-connect-agent`, and an interactive zsh starting while the socket is missing has prezto start a plain `ssh-agent` and repoint the symlink away from gpg-agent (findings 2 and 5).
- The always-enabled claim is cited to the GnuPG manual instead of to a one-time `lsof` on a specific pid (finding 7).

yqrashawn:
## pr-review-loop — pass 2

`89438e6aa88a` — **MERGEABLE**

Read-only reading. Verified by reading files on disk: the diff, `modules/yqrashawn/home-manager/dotfiles/default.nix`, `modules/yqrashawn/darwin/core.nix:109-111`, `modules/yqrashawn/home-manager/cli/prezto.nix:25,53`, plus installed artifacts (`/etc/zshenv`, the nix-darwin `set-environment`, prezto `modules/ssh/init.zsh`, gnupg 2.4.8 info manual, the current home-manager `activate`, `~/.gnupg/sshcontrol`). Nothing executed against the running gpg-agent, no build, no tests.

Closure of pass 1's targets: prose moved out of the hashed heredoc (now `default.nix:110-133`, heredoc at 139-147 is bare directives) — closed. The "ssh TTLs are inert" claim is gone and the replacement quote at 115-118 is verbatim from the GnuPG 2.4.8 manual (`gnupg.info-1:600-602`) — closed. `#167` (5a0fa1aa5) did add the ssh pair, so "simply missing until #167" is true. sshcontrol claim checked: 5 keygrip entries, every one TTL `0`, none `confirm`; manual says "A non-zero TTL overrides the global default" so "the global default governs" is right. Diff is comments-only (no non-comment line changes vs BASE), matching the PR body.

VERDICT[5ed1cefb]: MERGEABLE — 3 follow-ups to file

      [correctness/blocking]  none
      [correctness/followup]  none
      [coverage]              none
      [docs-accuracy]         3
      [style]                 none

1. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:158 — "only a new top-level login shell re-runs nix-darwin's gpg-connect-agent" is false; login-ness is not the gate. TRACED: `/etc/zshenv` is read by every zsh and sources `set-environment` under `[[ -o rcs ]]` plus `[ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]`; `set-environment:2` exports the var, `:21` runs `gpg-connect-agent updatestartuptty`, `:23` exports SSH_AUTH_SOCK. So a *non-login* interactive zsh whose ancestry lacks the var (shell spawned by a GUI-launched Emacs/kitty) does restart the agent, while a nested `zsh -l` does not — the comment understates recovery. Sibling, same class, same hunk's neighbour: line 152-153 states the opposite generically ("autostarted again ... by the next shell"); the two paragraphs contradict each other.
2. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:132 — "content-hashed, so a comment edit would change the store path and fire the onChange kill" names the wrong mechanism. TRACED in the installed `home-manager-generation/activate`: `_cmp` runs `cmp --quiet <new store file> ~/.gnupg/gpg-agent.conf` and sets `changedFiles[.gnupg/gpg-agent.conf]`, so the hook fires on *content* difference, not store-path difference. Conclusion still holds (heredoc comment text is content), but a store-path change with identical bytes would not fire it.
3. [docs-accuracy] modules/yqrashawn/home-manager/dotfiles/default.nix:160 — "An interactive zsh that starts while the socket is missing has prezto start a plain ssh-agent" overgeneralizes. TRACED through prezto `modules/ssh/init.zsh`: the plain-agent branch runs only when `[[ ! -S "$SSH_AUTH_SOCK" ]]` at `.zshrc` time, but in any shell that re-ran `set-environment`, `gpg-connect-agent` has already autostarted gpg-agent and recreated the socket before `.zshrc`; the branch is reachable only in shells that inherit the var and skip `set-environment`. Omission in the same sentence: when that branch does fire, prezto also `ssh-add`s the three identities from `prezto.nix:53-54` into the plain agent, so the drift is unlocked keys in a second agent, not just a repointed symlink.

PR #170, opened 2026-09-21
feat(pr-review): rotate a token pool instead of one pinned credential

yqrashawn:
## Why

The reviewer ran on **one** token (`~/Library/CloudStorage/Dropbox/sync/default-cc-token`). One token is one account's limit: when that account is spent, every review fails `MALFORMED (exited 1)` until it resets — on an org seat's monthly spend limit, days. That happened on 2026-09-20/21 and cost hours of review loop.

cchp already load-balances a `CLAUDE_TOKENS` pool for the sessions it spawns. This is that idea for the reviewer.

## What

New `pr-review.tokens`:

- **Pool**: `CLAUDE_TOKENS` from the reviewer's own environment (cchp passes it down to everything it spawns, so a review triggered from a session already has the live pool), else `CLAUDE_TOKENS` parsed out of cchp's `.env.local` — `PR_REVIEW_TOKENS_ENV_FILE` overrides the path — for reviews started from a terminal where nothing sourced it. The parser reads that file the way `source` does: comment lines skipped (so the several disabled pools in it stay disabled), last assignment wins, quotes and an unquoted trailing ` #` comment come off.
- **State in a file**, `$XDG_CACHE_HOME/pr-review-loop/tokens.edn`, under an flock — not an atom. Each review is its own babashka process, so nothing in memory survives to the next one and cchp's in-process approach has no equivalent here. Wall clock for the same reason: a monotonic reading means nothing to the process that reads the file next. **Tokens are never written there**, only a SHA-256 prefix.
- **Parking**: a limit banner (cchp's two `rate-limit-patterns`, copied with their self-match precaution) on a **non-zero exit** parks that token for an hour. Gated on the exit code deliberately — the reviewer reads repositories and quotes what it finds, and a review quoting a banner back would otherwise park a working token.
- With every token parked it still returns one, the soonest to come back. A park is a guess; refusing to run would turn that guess into a review that never happens, and a wrong guess costs one `MALFORMED`, which the ledger charges no pass for.

`reviewer.clj` gains `select-token` (pool → pinned file → nil, and nil still means "inherit the parent's auth") and a dynamic `*token*` bound once per `run!`. **That var is load-bearing, not tidiness**: selection is no longer idempotent, so `spawn-env` selecting separately from `run!` would put one token in the child and redact a *different* one — leaving the live credential in the PR comment. `CLAUDE_TOKENS` is also stripped from the child: the reviewer inherits the pool otherwise, and `redact` scrubs only the one token it was handed.

`default-token-file` keeps working and is now the fallback. Its docstring says so, and says what rotating costs it: reviewer spend no longer lands on one attributable account.

## Verification

- `bb test` — 290 tests, 816 assertions, **0 failures**.
- **Negative control**: re-armed the double-selection defect (made `spawn-env` re-select) → exactly `the-token-in-the-child-is-the-token-that-gets-redacted` failed, 1 failure. Restored from a copy → green again.
- **Live**: three consecutive `select-token` calls returned three distinct tokens, the cursor persisted as `{:parked {}, :cursor 3}`, and the state file held no token. With `CLAUDE_TOKENS` unset the `.env.local` path found the same three.

Not verified end to end: the **park** path has only fired in tests — no token was actually spent today. The banner patterns are copied verbatim from cchp's and unit-tested against both banner shapes.

## Summary by CodeRabbit

* **New Features**
  * Added rotating credential-pool support for code reviews.
  * Automatically skips credentials that reach usage limits and reuses them after a cooldown period.
  * Supports loading credentials from environment variables or local environment files.

* **Bug Fixes**
  * Ensured a single credential is used consistently throughout each review.
  * Prevented credential-pool variables from being passed to spawned processes.
  * Improved malformed review-result diagnostics.

* **Tests**
  * Added coverage for credential rotation, fallback behavior, limit handling, persistence, redaction, and environment isolation.


yqrashawn:
## pr-review-loop — pass 1

`bed14f845f30` — **NOT MERGEABLE**

Verified by RUNNING: findings 1, 2, 3, 4 and 6 (bb probes in the throwaway worktree — negative control on the strip, parser probes, banner probes, one `reviewer_test` var under a scratch `XDG_CACHE_HOME`). Finding 5 is a reading of the source plus a hypothesised trigger; finding 7 is a reading of the diff. `bb -e` ran `pr-review.tokens-test` alone (11 tests, 30 assertions, 0 failures); full suite not run.

VERDICT[015bcfce]: NOT MERGEABLE — `the-pool-does-not-reach-the-child` certifies pool isolation but passes with the isolation removed whenever `CLAUDE_TOKENS` is unset in the test process

  [correctness/blocking]  none
  [correctness/followup]  4
  [coverage]              1
  [docs-accuracy]         1
  [style]                 1

1. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:154 — `the-pool-does-not-reach-the-child` asserts the child's `$CLAUDE_TOKENS` is `[]`, but the child inherits the TEST process's environment; `tokens/pool` is redefed and never reaches the env. TRACED and RUN: deleted `"CLAUDE_TOKENS" nil` from `spawn-env`, ran this var alone — green with `CLAUDE_TOKENS` unset (a plain shell here has it unset), `(not (= "[]" "[real-a,real-b]"))` only with it set. The one assertion guarding credential-pool isolation is environment-dependent. Assert on the map instead: `(contains? (spawn-env) "CLAUDE_TOKENS")` and its value `nil`, or set the var in the spawn's own parent env.
2. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:56 — `unquote-value` handles quotes OR a trailing ` #` comment, never both. RUN: `CLAUDE_TOKENS="aaa,bbb" # jason, 0g` parses to `["\"aaa" "bbb\""]`. `.env.local` is documented at tokens.clj:36 as carrying comments on these lines, and the fixture at tokens_test.clj:40 shows that comment style. Worse than junk: `pool` is non-empty, so `select!` returns a token and `(or (tokens/select!) (oauth-token))` at reviewer.clj:222 short-circuits — the pinned-file fallback is skipped and every review runs on a quote-mangled credential. Sibling: `'aaa,bbb' # c` same shape. Trigger: the operator quotes the value.
3. [correctness/followup] claude-code-plugins/test/pr_review/reviewer_test.clj:115 — `run!` with `:spawn-fn` but no `:token-fn` and no `tokens/pool` redef now calls the real `tokens/select!`: it reads the operator's live pool and writes the live cursor. RUN: that var alone under a scratch `XDG_CACHE_HOME` left `{:parked {}, :cursor 1}` on disk. Siblings at :140 and :146 — `bb test` advances the shared cursor by 3. `tmp-state`'s own docstring (tokens_test.clj:8) names this hazard. No park is reachable from these stubs (all exit 0 or throw before the gate), so the cost is bounded to skipped tokens.
4. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:247 — `(?:\w+ )?` does not match a hyphenated qualifier. RUN: `"You've hit your 5-hour limit · resets 3am ("` is not matched; `"...usage limit..."` is. TRACED that the regex rejects it; INFERRED, not traced, that the CLI emits that wording — the banner set is copied from cchp and I cannot read cchp from here. If it does, the spent token is never parked and the next rotation hands it out again.
5. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:259 — `redact` scrubs only the ONE selected token, but the pool is now N tokens and `spawn-env`'s env-strip does not cover the file they live in. TRACED: `denied-tools` (reviewer.clj:109-115) permits Read/Grep/Glob/Bash, and tokens.clj:41 puts the literal `.env.local` path into the diff the reviewer is told to read. HYPOTHESISED trigger: the reviewer `cat`s that path while checking the parser — then the whole pool lands unredacted in the PR comment. Cheap close: scrub every `tokens/pool` entry, not just the selected one.
6. [docs-accuracy] claude-code-plugins/hooks/pr_review/reviewer.clj:243 — "CLAUDE_TOKENS is stripped from the child" contradicts the inline comment 13 lines below and the measured behaviour. RUN: with `:extra-env {"CLAUDE_TOKENS" nil}` the child's `env | grep -c CLAUDE_TOKENS` is `1` — present and empty, not stripped.
7. [style] claude-code-plugins/hooks/pr_review/reviewer.clj:435 — `parse-counts`' binding alignment and `parse-output`'s whole `if` block (reviewer.clj:489-512) are re-indented with no behaviour change, in a PR about token rotation. Unrelated churn in the diff the next reviewer has to read past.

yqrashawn:
## pr-review-loop — pass 2

`04567fa113d8` — **MERGEABLE**

Verified by RUNNING (bb probes in the throwaway worktree): pass-1 closure for findings 1–4, and new finding 1. Reading of the source: findings 2 and 3. Ran `pr-review.tokens-test` + `pr-review.reviewer-test` (55 tests, 176 assertions) and `pr-review.trigger-test` (68 tests, 211 assertions) under a scratch `XDG_CACHE_HOME` — all green, and no rotation state was written there, which is the evidence the hermetic fixture holds. Full `bb test` not run.

Closure of pass 1:

```
1. [coverage] tokens_test.clj:154 — environment-dependent pool-isolation assertion
2. [correctness/followup] tokens.clj:56 — unquote-value handles quotes OR a comment, never both
3. [correctness/followup] reviewer_test.clj:115 — run! advances the operator's live cursor
4. [correctness/followup] tokens.clj:247 — hyphenated qualifier unmatched
5. [correctness/followup] reviewer.clj:259 — redact scrubs only the selected token
6. [docs-accuracy] reviewer.clj:243 — "stripped from the child"
7. [style] reviewer.clj:435 — unrelated re-indentation churn
```

1 closed — `tokens_test.clj:185` now asserts on the map (`contains?` + `nil?`), not on a child's stdout. 2 closed — measured `"a,b" # jason` → `["a" "b"]`, `'c,d' # p` → `["c" "d"]`, `e,f # j` → `["e" "f"]`. 3 closed — `test_env/hermetic-tokens` redefs `state-path`; with `XDG_CACHE_HOME` pointed at a scratch dir, running reviewer-test and trigger-test created nothing under it. 4 closed — `5-hour` now matches. 5 **partially** closed: the pool is scrubbed, the pinned fallback is not (finding 1 below). 6 closed — docstring now says BLANKED; I measured `env | grep -c '^CLAUDE_TOKENS='` = 1 with `:extra-env {"CLAUDE_TOKENS" nil}`, so the new wording is accurate. 7 not closed, and widened.

VERDICT[af043df0]: MERGEABLE — 3 follow-ups to file; no blocking finding, and no test certifying a safety property it does not check

  [correctness/blocking]  none
  [correctness/followup]  1
  [coverage]              none
  [docs-accuracy]         1
  [style]                 1

1. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:282 — `redact`'s `secrets` is `pool ++ [selected]`; the pinned `default-token-file` credential is in neither whenever the pool is non-empty, so it is the one credential this process knows about that is no longer scrubbed. REGRESSION, not pre-existing: before rotation `token` was always the pinned token, so it was always in the marker set. RUN, in the worktree: with `tokens/pool` redefed to two tokens and `default-token-file` pointed at a file holding `sk-ant-oat01-PINNED-SECRET`, `run!`'s `:out` came back `"cat default-cc-token: sk-ant-oat01-PINNED-SECRET ; and pool: [redacted]"` — the pool member scrubbed, the pinned one verbatim. Reachability is the same shape pass 1 gave for finding 5 and is HYPOTHESISED, not traced: `denied-tools` (reviewer.clj:115-130) grants Read/Grep/Glob/Bash and reviewer.clj:175-176 constructs the path in source the reviewer is routinely asked to read. Narrower than the pool case — the pinned token is in the child's env only when the pool is EMPTY, and there it is scrubbed — so a file read is the only route. CodeRabbit's inline prompt on this PR asked for "every credential in the pool used by the run, **including the fallback token**"; the fallback half was not done. One-line fix: `conj` `(oauth-token)` into `secrets`. Same-class sibling, smaller: `redact` re-reads `tokens/pool` minutes after selection, so a pool edited mid-review leaves the departed members unmarked — the selected token is still covered via the explicit `conj`.
2. [docs-accuracy] claude-code-plugins/test/pr_review/test_env.clj:13 — the namespace docstring says "The seam is `tokens/state-path` and `tokens/pool`, so the default is closed HERE", which the function's own docstring at :26 contradicts ("`state-path` ONLY") and the body at :35 contradicts (`with-redefs [tokens/state-path …]`, nothing else). A reader taking the ns docstring at its word believes fixtured tests see no pool; TRACED — `trigger_test`'s `review-opts` (trigger_test.clj:96-116) passes no `:token-fn`, so `run!` calls the real `select-token`, which reads the operator's live `CLAUDE_TOKENS`/`.env.local`. Harmless as implemented (reads only; the cursor is scratch) but the two docstrings cannot both be right.
3. [style] claude-code-plugins/hooks/pr_review/reviewer.clj:463 — whitespace-only re-indentation, again and wider than pass 1: `parse-counts`' `n  ` alignment here, the whole `parse-output` `if` block at :510-543, and `tokens_test.clj:129`. Verified by running `git diff --no-ext-diff -w bed14f845f30..HEAD` on reviewer.clj — 71 lines with `-w` against 112 without, and zero non-whitespace change in either region. Unrelated churn in a PR about token rotation, and the fix commit added to it rather than dropping it.

yqrashawn:
## Review loop summary — 2 passes

|---|---|---|---|---|---|---|
| 1 | NOT MERGEABLE | 0 | 4 | 1 | 1 | 1 |
| 2 | MERGEABLE | 0 | 1 | 0 | 1 | 1 |

(Pass 1 also cost one `MALFORMED` attempt before it: the reviewer's account had hit its org monthly spend limit — the exact failure this PR exists to fix. That attempt wrote no ledger row, and the retry ran on the next pool token. The park is visible in `~/.cache/pr-review-loop/tokens.edn`: the spent account's hash parked, the cursor moved on. That is the end-to-end verification I could not do before opening this.)

### Fixed on this PR (pass 1 → pass 2, all confirmed closed by the reviewer)

Each has a negative control — the defect was re-armed and the named test watched to fail, then restored.

1. **[coverage]** `the-pool-does-not-reach-the-child` could not fail: it expected `[]` from a child, which is also what the child prints with the strip *deleted* whenever the test process has no `CLAUDE_TOKENS` — every plain shell. Now asserts on `spawn-env`'s map.
2. **[correctness]** `unquote-value` handled quotes **or** a trailing comment, never both, so `CLAUDE_TOKENS="a,b" # jason` gave a pool of quote-mangled credentials — worse than none, because non-empty short-circuits the pinned-file fallback.
3. **[correctness]** Every test reaching `run!` without pinning a token advanced the operator's **live** rotation cursor. Closed at the seam (`test-env/hermetic-tokens`), not at the call sites; verified by the real state file being byte-identical across a full `bb test`.
4. **[correctness]** `(?:\w+ )?` missed a hyphenated window (`5-hour`), so that banner would never park. Widened to `[\w-]+` — a deliberate divergence from cchp's copy, recorded with the asymmetry that decides it.
5. **[correctness]** `redact` scrubbed only the selected token, leaving the rest of the pool printable. Now scrubs the whole pool. *(Pass 2 found the other half of this — see below.)*
6. **[docs-accuracy]** "stripped from the child" contradicted the measurement 13 lines below it. It is blanked, not removed.

### Deferred

- **[correctness/followup] `redact` does not scrub the pinned `default-token-file` token when the pool is non-empty.** Real, and a regression from this PR — before rotation the selected token *was* the pinned one. Follow-up PR, per the loop's rules.
- **[docs-accuracy] `test_env`'s namespace docstring** still says the fixture closes `pool` as well; the function docstring and body say `state-path` only. Correct finding — carried into the same follow-up PR, since it is one paragraph in a file that PR touches.
- **[style] whitespace churn.** Dropping this one. Pass 1's instance was real and was fixed by restoring both forms to the base version; pass 2 re-raised it from a `bed14f845..HEAD` diff, which is the *fix commit* and so is whitespace-only by construction. Against the merge base the PR has no churn: `git diff origin/master...HEAD` and `git diff -w origin/master...HEAD` both report `101 insertions(+), 18 deletions(-)` for `reviewer.clj`.

### Verification

`bb test` — 292 tests, 826 assertions, 0 failures. `clj-kondo` — 0 errors. Mergify/CodeRabbit checks green, `mergeStateStatus: CLEAN`.

PR #171, opened 2026-09-21
fix(pr-review): scrub the pinned token too, and fix the fixture's docstring

yqrashawn:
Follow-up to #170, from its pass-2 findings.

## The pinned token was the one credential this process did not scrub

A **regression from rotation**, not a pre-existing gap: before it, the selected token always *was* the pinned `default-token-file` token, so it was always in the marker set. With a non-empty pool it is in neither the pool nor the selection. Reachable the same way the pool was — the reviewer has Read, Grep, Glob and Bash, and `reviewer.clj` constructs that path in source it is routinely asked to read.

Fixed by moving the **question**, not by adding a third `conj` at the sink: `credentials` now owns *which* secrets are at risk, `redact` takes the collection it returns, and `run!` computes it once. Twice now that question has been answered wrong by being answered at the call site — once for the pool (#170 pass 1), once for the pinned file (pass 2) — which is the tell that the call site is the wrong place for it.

Computing it once also closes the sibling the reviewer named in the same finding: `redact` used to re-read `tokens/pool` at each sink, minutes after selection, so a pool edited mid-review left the departed member verbatim in the output of the review still running on it.

## And the fixture's docstring

`test-env`'s *namespace* docstring claimed the fixture closes `tokens/pool` as well, which its own function docstring and body contradict. It closes `state-path` only, and a fixtured test does still read the live pool — harmless, since that read mutates nothing and the cursor is scratch. Now it says that.

## Verification

Both fixes have a negative control — the defect re-armed, the named test watched to fail, then restored:

- dropping `(oauth-token)` from `credentials` → `the-pinned-fallback-token-is-redacted-even-when-the-pool-is-used` fails
- reading the secrets at each sink instead of once → `a-pool-edited-mid-review-does-not-unmark-the-token-in-flight` fails

`bb test` — 294 tests, 829 assertions, 0 failures. `clj-kondo` — 0 errors.


yqrashawn:
## pr-review-loop — pass 1

`46f1f7cbe916` — **MERGEABLE**

Verified by RUNNING: both new tests pass, and both negative controls reproduce (dropping `(oauth-token)` from `credentials` fails `the-pinned-fallback-token-is-redacted-even-when-the-pool-is-used`; inlining `(credentials token)` into the `scrub` lambda fails `a-pool-edited-mid-review-does-not-unmark-the-token-in-flight`). `clj-kondo` 0 errors. Worktree restored clean after. Everything else below is a reading of the source.

VERDICT[90566fc4]: MERGEABLE — 3 follow-ups to file

  [correctness/blocking]  none
  [coverage]              none
  [style]                 none

1. [correctness/followup] claude-code-plugins/hooks/pr_review/reviewer.clj:282 — `credentials` covers only the pool `tokens/pool` resolves to, and `pool` skips comment lines (tokens.clj:102) and prefers the environment over the file (tokens.clj:135-138). Two sets of real credentials in `.env.local` therefore stay in neither `secrets` nor the marker set: (a) the "several commented-out CLAUDE_TOKENS lines" tokens.clj:36-37 says that file holds — prior pools, i.e. live tokens for real accounts; (b) the file's live line when the inherited `CLAUDE_TOKENS` env var wins and the operator has since swapped pools — the same mid-review edit the new test models, seen from the other side. Reachability is exactly the argument `credentials`' own bullet 2 makes: `tokens/default-env-file` builds that path in source the reviewer is routinely asked to read (tokens.clj:40-41), and a `cat` prints all of them verbatim into the PR comment. Case (a) needs only the `cat`; case (b) needs the `cat` plus a mid-session pool swap. Traced through the source; the claim that comment lines carry real tokens rests on tokens.clj:36-37, not on a read of the operator's file. Not blocking: the diff delivers what it claims, `redact`'s docstring (reviewer.clj:304-306) explicitly disclaims values the process does not know, and this is a scope boundary rather than a regression — but it is the unswept sibling of the leak this PR and #170 both chased.

2. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:186 — `the-pool-does-not-reach-the-child`'s docstring still says "`redact` scrubs only the ONE token it was given, so every other token in the pool would be unredacted". False since #170 and doubly so now: `redact` takes a collection and `credentials` supplies pool + selection + pinned. Untouched by this diff, but it is the same class of stale-rotation docstring the PR fixes in `test-env`.

3. [docs-accuracy] claude-code-plugins/hooks/pr_review/reviewer.clj:279 — "Too-short values are dropped here rather than at each use" is contradicted by reviewer.clj:311, where `redact` filters `(>= (count %) 8)` again. The invariant lives in two places; only one claims to own it. (`redact`'s own test at reviewer_test.clj:561 exercises the second filter, so the duplication is deliberate — the docstring is what is wrong.)

yqrashawn:
## Review loop summary — 1 pass

|---|---|---|---|---|---|---|
| 1 | MERGEABLE | 0 | 1 | 0 | 2 | 0 |

The reviewer reproduced both negative controls itself, which is the check that mattered here: this PR is two safety properties, and a test that cannot fail would have certified them falsely.

### Carried into a follow-up PR

- **[correctness/followup] the commented-out pools in `.env.local` are real credentials and are not in the marker set.** `tokens/pool` skips comment lines and prefers the environment over the file, so the prior pools that file is documented as holding — live tokens for real accounts — reach `redact` from neither direction. Reachability is the same argument `credentials`' own docstring makes for the live pool: the path is built in source the reviewer is asked to read, and one `cat` prints all of them. Correct, and it is the unswept sibling of the leak this PR and #170 both chased, so it gets fixed rather than dropped.
- **[docs-accuracy] two stale rotation docstrings** — `the-pool-does-not-reach-the-child` still says `redact` scrubs only one token, and `credentials` claims to own the too-short filter that `redact` also applies. Both fold into the same follow-up, since it touches both files.

### Verification

`bb test` — 294 tests, 829 assertions, 0 failures. `clj-kondo` — 0 errors.

PR #172, opened 2026-09-21
fix(pr-review): a disabled pool in the env file is still a live credential

yqrashawn:
Follow-up to #171, from its pass-1 `[correctness/followup]` finding.

## A disabled pool is still a live credential

`credentials` drew its marker set from `tokens/pool`, which skips comment lines and prefers the environment over the file. Both are right for the question `pool` answers — *which token do I authenticate with*. Both are wrong for the other question: *which strings, if the reviewer printed them, would be a leaked credential.*

Two sets of real tokens fell in that gap:

- the **commented-out `CLAUDE_TOKENS` lines** the env file is documented as carrying — those are prior pools, i.e. live tokens for real accounts;
- the file's **live line** on a run where the inherited environment won instead.

One `cat` of a path `tokens.clj` names in source the reviewer is routinely asked to read prints every one of them, which is exactly the reachability argument `credentials`' own docstring already makes for the live pool.

`known-tokens` answers the second question: every `CLAUDE_TOKENS` assignment in the file, `#` or not, plus the environment's. Nothing authenticates with it and its order is meaningless. **`pool` is unchanged** — a disabled line must still never be logged in with.

The explicit-path arity is the file **alone**, for the same reason `pool`'s `:env-file` is: an ambient variable winning over a named file makes the argument a no-op wherever it is set — which is every reviewer cchp spawns, and every run of these tests. Caught by the tests, which read the operator's live pool through it before that was fixed.

## Also

The two stale docstrings the same pass named: the child-isolation test no longer describes `redact` as single-token, and `credentials` no longer claims to be the sole owner of the 8-character floor that `redact` also applies.

## Verification

Negative control: drawing from `tokens/pool` again fails six assertions across three tests, including `a-retired-token-the-reviewer-prints-is-redacted`.

Two defects in the new test itself were caught and are recorded in its comments — a fixture read before it was written (which made the assertion pass on an empty marker set) and a lazy stub that called the var it was stubbing (a `StackOverflowError`, which `credentials`' `catch Exception` does not catch).

`bb test` — 296 tests, 833 assertions, 0 failures. `clj-kondo` — 0 errors.


yqrashawn:
## pr-review-loop — pass 1

`a7e0cee9a611` — **MERGEABLE**

VERDICT[bba06043]: MERGEABLE — 5 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  3
  [coverage]              1
  [docs-accuracy]         1
  [style]                 1

1. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:170 — the no-arg arity scans `PR_REVIEW_TOKENS_ENV_FILE` **or** `default-env-file`, never both, while `default-env-file`'s path stays a literal in source at tokens.clj:40-41. The PR's own reachability argument is "the reviewer cats a path this source names"; with the override set (to another file, or to `""`, which `or` treats as set exactly as `pool` line 136 does) the default file's live and disabled pools are back outside the marker set and a `cat` of the hardcoded path prints them. TRACED in source; nothing in the repo sets the variable (grep, repo-wide), so the trigger is manual operator use only — hence followup.
2. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:189 — the marker set is keyed on `(= "CLAUDE_TOKENS" (str/trim k))` after one `^#+\s*` strip, so plausible file shapes fall outside it. RUN, not inferred: a probe file gave `["tok-retired-aaaa" … "tok-live-eeee"]` and silently dropped `# # CLAUDE_TOKENS=tok-retired-ffff`, `# set CLAUDE_TOKENS=tok-prose-gggg` and `# OLD_CLAUDE_TOKENS=tok-old-hhhh` (`#  export …` and `## …` are correctly caught). Sibling, INFERRED: the same one-`cat` argument covers every other assignment in cchp's `.env.local` — only `CLAUDE_TOKENS`-keyed lines are marked and any other credential in that file is printed verbatim. I did not read the operator's `.env.local`, so the second half is a hypothesis about its contents, not a trace.
3. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:192 — new over-redaction direction: commented-out lines are now markers, and commented-out lines are exactly where a placeholder lives. A `# CLAUDE_TOKENS=REPLACE_ME_WITH_TOKEN` line yields a ≥8-char marker that `redact` then substitutes everywhere in `:out`, `:err` and the findings file, garbling review text that has nothing to do with a credential. INFERRED — depends on the operator's file carrying such a line; the mechanism itself is traced (tokens.clj:192-195 → reviewer.clj:287 → reviewer.clj:314).
4. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:252 — the exact-set assertion is what enforces "an explicit `path` is the file alone", and it can only fail where `CLAUDE_TOKENS` is set in the process environment. RUN: `env -u CLAUDE_TOKENS PR_REVIEW_TOKENS_ENV_FILE=/no/such/file bb` → 17 tests, 46 assertions, 0 failures, and the property is unexercised in that run. Same gap one level up: no test exercises the no-arg arity at all (env ∪ file), and every `run!`-level test stubs `tokens/known-tokens` (tokens_test.clj:275, :299, :317), so the `credentials` → real `known-tokens` seam has no coverage. This does NOT flip the verdict: the claim the test's own docstring makes — disabled lines belong to the marker set — does fail when broken, which I confirmed by negative control.
5. [docs-accuracy] claude-code-plugins/hooks/pr_review/tokens.clj:159 — "So: every assignment in the file, commented or not" overstates what line 189 does twice over: only `CLAUDE_TOKENS`-keyed assignments, and not every spelling of those (finding 2). A reader checking the leak argument against this sentence concludes the file is covered.
6. [style] claude-code-plugins/hooks/pr_review/tokens.clj:146 — `known-tokens` is placed under the `;;; Cross-process rotation state` header (line 144) but has nothing to do with the cursor or the parked set; it belongs with `;;; Where the pool comes from` (line 54) beside `env-file-var` and `pool`, whose parsing it duplicates.

What I RAN, in the throwaway worktree: a probe of `known-tokens` over a synthetic env file (finding 2, and confirming the explicit arity ignores an ambient `CLAUDE_TOKENS` while the no-arg arity includes it); `pr-review.tokens-test` in isolation, green at 17/46 (finding 4); and the PR body's negative control — reverting `credentials` to `tokens/pool` produced exactly 6 failures across 3 tests including `a-retired-token-the-reviewer-prints-is-redacted`, matching the claim, and the edit was reverted (`git status` clean). Everything else is a reading of the source. No probe touched the operator's real `.env.local` or rotation state.

`pool` is genuinely unchanged, and the new marker set is a strict superset of what `pool` returned in all three precedence cases, so redaction does not regress. Findings 1-3 are all the same root — the marker set is derived from a narrower rule than the "one `cat` of a named path" threat it is justified by — and are worth fixing together.

yqrashawn:
## Review loop summary — 1 pass

|---|---|---|---|---|---|---|
| 1 | MERGEABLE | 0 | 3 | 1 | 1 | 1 |

Merging. The reviewer confirmed the negative control itself and confirmed `pool` is unchanged and the new marker set is a strict superset in all three precedence cases, so redaction does not regress.

## Not fixed here — and not in a fourth PR either

The reviewer's own note is the important one: **findings 1–3 are all the same root** — the marker set is derived from a narrower rule than the "one `cat` of a named path" threat it is justified by. So is this PR. So was #171's finding. So was #170's.

That is three consecutive passes, on three PRs, each finding the next gap one enumeration deeper:

| pass | what was outside the marker set |
|---|---|
| #170 p1 | the rest of the live pool |
| #171 p1 | the pinned `default-token-file` token |
| #172 p1 | the disabled pools; a non-default `PR_REVIEW_TOKENS_ENV_FILE`; `# # CLAUDE_TOKENS=`, `# set CLAUDE_TOKENS=`, `# OLD_CLAUDE_TOKENS=`; every non-`CLAUDE_TOKENS` credential in the same file |

When the same function draws a finding on consecutive passes, the validation **strategy** is the defect, not the input. Enumerating which strings are credentials cannot converge — each pass names one more spelling — and finding 3 shows the cost of widening it blindly: a `# CLAUDE_TOKENS=REPLACE_ME_WITH_TOKEN` placeholder becomes a marker that garbles unrelated review text.

The strategy question is whether to keep enumerating sources at all, or to redact by **shape** (an `sk-ant-oat01-…`-shaped string in the output is a credential, wherever it came from) — which is a reversal of `redact`'s stated "a value we do not know cannot be matched" and is the operator's call, not mine to make inside a review fix.

Filed as an issue with all six findings rather than opened as a fourth PR. Findings 4 (coverage), 5 (docs) and 6 (style) are carried there too — 4 is a genuinely environment-dependent assertion, of the same kind #170 pass 1 caught, and it belongs with whatever replaces the enumeration.

## Verification

`bb test` — 296 tests, 833 assertions, 0 failures. `clj-kondo` — 0 errors.

Issue #173, opened 2026-09-21
pr-review: redaction enumerates credential sources, and cannot converge that way


yqrashawn:
Three consecutive review passes, on three PRs, each found the next gap one enumeration deeper in the same function:

| pass | what was outside the marker set |
|---|---|
| #170 p1 | the rest of the live pool |
| #171 p1 | the pinned `default-token-file` token |
| #172 p1 | the disabled pools; a non-default `PR_REVIEW_TOKENS_ENV_FILE`; `# # CLAUDE_TOKENS=`, `# set CLAUDE_TOKENS=`, `# OLD_CLAUDE_TOKENS=`; every non-`CLAUDE_TOKENS` credential in the same file |

Each was fixed by adding one more source to `reviewer/credentials`. That is the tell: the **strategy** is the defect, not the input. Enumerating which strings are credentials cannot converge, because each pass names one more spelling, and widening it blindly has a cost in the other direction — see finding 3 below.

## The threat, stated once

The reviewer runs with `Read`, `Grep`, `Glob` and `Bash` under `permissions.defaultMode: bypassPermissions`. Everything it prints is republished three ways: `:out` becomes the PR comment and the findings file, `:err` is streamed to a file the author reads. `tokens.clj` and `reviewer.clj` both build credential file paths as literals in source the reviewer is routinely asked to read. **One `cat` of either path puts whatever is in that file into a public PR comment.**

`credentials` is the only thing standing between that and the comment.

## Open findings from #172 pass 1

1. **[correctness]** The no-arg `known-tokens` scans `PR_REVIEW_TOKENS_ENV_FILE` **or** `default-env-file`, never both — and `or` treats `PR_REVIEW_TOKENS_ENV_FILE=""` as set, exactly as `pool` does. With the override pointed anywhere else, the default file's pools are outside the marker set while its path is still a literal in source.
2. **[correctness]** The marker set is keyed on `(= "CLAUDE_TOKENS" (str/trim k))` after one `^#+\s*` strip. Measured as dropped: `# # CLAUDE_TOKENS=…`, `# set CLAUDE_TOKENS=…`, `# OLD_CLAUDE_TOKENS=…`. Sibling, untraced: every **other** credential in cchp's `.env.local` — that file holds far more than Claude tokens, and only `CLAUDE_TOKENS`-keyed lines are marked.
[...]

PR #174, opened 2026-09-22
feat(pr-review): park a spent token until the reset its banner states

yqrashawn:
Answers a question the operator asked of #170: *does the reviewer record limit-reached and pause that token for a while?* It does — and "a while" was measurably the wrong length.

## The measurement

Yesterday an org seat hit its limit mid-review. The park worked: the account was recorded at **22:24** and skipped, and the retry ran on the next token and produced the review. But the park was a flat hour, so the account came back into rotation at **23:24** — and its banner said `resets 3am (Asia/Shanghai)`. Verified this morning with a one-token probe: it was live again at 03:00, not 23:24.

That leaves a 3h36m window where any review drawing that token fails, and **the failure is not retried**: a `MALFORMED` attempt writes no ledger row and the review simply does not happen until someone pushes again.

## The fix

One hour was a guess made because the structured `rate_limit_event` cchp's token-manager reads is not available to a `claude -p`. But the prose states the reset, in three shapes all seen live — a bare `3am`, minutes as in `8:30pm`, and a dated `Sep 15 at 3am`. `reset-at-ms` parses it in the zone the CLI printed, and `park!` runs the park to that instant. No parse — no banner, or prose that cannot be placed on a clock — still means the hour.

Undated means the **next occurrence** of that time, which for a weekly window is a lower bound. A lower bound is the right error to make: it costs one retry to discover, where over-parking is idle capacity that nothing corrects.

Two bounds, both borrowed from cchp's token-manager with its reasoning:

- an instant more than **8 days** out is *disbelieved*, not clamped — the number comes from another machine's clock, and clamping would still park for 8 days on a value already known to be wrong;
- a park **never shortens** an existing one — two parks mean two rejections and the later instant is what both agree on, or a banner-less second rejection replaces a parsed multi-day park with an hour.

## Verification

Three negative controls, each armed and watched to fail:

| armed defect | test that failed |
|---|---|
| ignore the parsed reset | `a-park-lasts-until-the-stated-reset` |
| let a park shorten | `a-second-park-never-shortens-the-first` |
| do not pass the banner from `run!` | `the-banner-reaches-park-through-run` |

The shortening one is asserted on the **stored instant**, not through `select!`, and the comment says why: its first version went through `select!` and **passed with the bug armed** — with both tokens available `select!` returns whichever the cursor is on. Its second version then failed for a reason that was not the bug, because a `select!` after the reset prunes the park and leaves nothing for a later one to shorten.

Parser probed against `12am`/`12pm` (the ends of the 12-hour clock), a foreign zone, a bad zone, a date beyond the bound, and a non-hour.

`bb test` — 300 tests, 852 assertions, 0 failures. `clj-kondo` — 0 errors.

Does not touch the open redaction question in #173.

## Summary by CodeRabbit

* **Bug Fixes**
  * Token limits are now paused until the reset time reported by the service, rather than using a fixed one-hour delay.
  * Reset times are interpreted across time zones and date formats for more accurate scheduling.
  * Existing longer pauses are preserved and are not shortened by subsequent limit notices.

* **Tests**
  * Added coverage for reset-time parsing, time zones, date formats, and token pause behavior.


yqrashawn:
## pr-review-loop — pass 1

`80b1ed2ff29a` — **NOT MERGEABLE**

Ran: `bb` tokens-test namespace (21 tests, 65 assertions, 0 failures), plus three probe scripts against the real namespace (parser inputs, an armed drop-the-banner defect, and a mixed-output banner). Findings 1 and 3 were measured by running; findings 2, 4, 5, 6 are a reading of the source.

VERDICT[09000962]: NOT MERGEABLE — `reset-at-ms` takes the FIRST `resets …` in the reviewer's whole output, so prose the reviewer quoted wins over the real banner and parks a live account for up to 8 days (measured: 148.6 h instead of 4.6 h)

  [correctness/blocking]  1 findings
  [coverage]              1 findings

1. [correctness/blocking] claude-code-plugins/hooks/pr_review/tokens.clj:372 — `(re-find reset-re banner)` takes the FIRST match in the string, and `reviewer.clj:366` hands it `(str (:out res) "\n" (:err res))` — the whole review body, not the banner. TRACED: the `limited?` gate at `reviewer.clj:362-363` only requires exit≠0 plus banner-shaped text *anywhere* in out or err, and the comment right above it (`reviewer.clj:358-361`) already concedes the reachable case — "the reviewer reads repositories and quotes what it finds". MEASURED with `bb`: output whose review text quotes `resets Sep 28 at 3am (Asia/Shanghai)` above a real `· resets 3am (Asia/Shanghai)` banner parses to a **148.6 h** park where the true banner says **4.6 h**. Sibling, same root: `limited?` is checked on `:out` and `:err` separately but the parse runs on the concatenation, so a quoted string in `:out` overrides a genuine banner in `:err`. Compounding: `park!:450`'s never-shorten `max` plus the absence of any unpark means the bad instant cannot be corrected — a later, correct rejection is discarded. Fix is local: parse from the same match `limited?` found, or take the last match rather than the first.

2. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:450 — the never-shorten `max` makes any over-long park permanent for its full duration; there is no unpark path anywhere in the namespace (grep: only `park!` and `prune` write `:parked`, and `prune` only drops *expired* entries). Trigger independent of finding 1: the ns docstring at :15-20 states the clock is WALL CLOCK and accepts that "an NTP step or a suspended host mis-times a park" — that acceptance was priced at one hour; a forward clock step during a parsed park now writes an instant up to `max-park-ms` (8 days) out that no later rejection can shorten. Degraded rather than fatal, because `choose:250-263` still returns the earliest-expiring token when all are parked.

3. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:169 — the `testing` string claims "the two ends of the 12-hour clock, which is where an am/pm conversion goes wrong", but the block asserts `8:30pm` (the `else` branch) and `12am` twice (:174, :177); nothing exercises the `(= 12 h12) → 12` branch at `tokens.clj:377`. If 12pm mapped to hour 0 the suite stays green. VERIFIED by running that the branch is currently correct (`resets 12pm` → 12:00), so this is an untested branch, not a live bug. The PR body repeats the unmet claim: "Parser probed against `12am`/`12pm`".

4. [docs-accuracy] claude-code-plugins/hooks/pr_review/tokens.clj:19 — five places still assert a park is always one hour, which this PR makes false, and one of them is load-bearing for a safety argument. Siblings, reported together: :19 "A park is a one-hour guess either way, and the next rejection re-parks it" (both halves now false — the never-shorten rule at :450 means the next rejection *cannot* re-park it shorter); :44 "How long a token is skipped after it answered with a limit banner: 1 hour"; :48-49 "there is no `resetsAt` to believe"; :251 "A park is a guess (see `park-ms`)"; :311 "a pattern one word too wide costs at most one token parked for an hour" — the bound is now 8 days, which is exactly the cost in finding 1; `reviewer.clj:361` "would otherwise park a working token for an hour"; and `test/pr_review/tokens_test.clj:115` "a park is a one-hour GUESS — the CLI's prose banner carries no reset this can parse".

5. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:180 — "and the previous midnight for a caller a day earlier" describes the assertion backwards: `now` is 2026-09-20 13:00 and the expected value is 2026-09-21 00:00, the NEXT midnight. `reset-at-ms` filters on `.isAfter now` (:409) and can never return a past instant, so the message tells a reader the opposite of the contract.

6. [style] claude-code-plugins/hooks/pr_review/tokens.clj:387 — `clojure.string/upper-case` and `clojure.string/lower-case` written fully qualified inside a namespace that aliases `[clojure.string :as str]` and uses `str/` everywhere else.

yqrashawn:
## pr-review-loop — pass 2

`8b7ede32932e` — **MERGEABLE**

Verified pass-1 findings against current source; ran the tokens-test namespace and three probe scripts.

**Closure of pass 1**

```
1. [correctness/blocking] tokens.clj:372 — re-find takes the FIRST match
```
Closed at `claude-code-plugins/hooks/pr_review/tokens.clj:450` — `re-seq` + `keep` + `sort` + `first`. MEASURED: quoted `Sep 28 at 3am` above a genuine `resets 3am` now yields 4.60 h, the banner's own instant, in both orderings. Residual path below as finding 1.

```
2. [correctness/followup] tokens.clj:450 — never-shorten max is permanent
```
Not closed; now acknowledged in the ns docstring at `tokens.clj:23` as Issue #175. Re-listed as finding 3.

```
3. [coverage] tokens_test.clj:169 — 12pm branch untested
```
Closed — assertion at `test/pr_review/tokens_test.clj:187`; `12pm` → 2026-09-22T12:00.

```
4. [docs-accuracy] tokens.clj:19 — five places assert a one-hour park
5. [docs-accuracy] tokens_test.clj:180 — "previous midnight" backwards
6. [style] tokens.clj:387 — fully-qualified clojure.string
```
All closed: `tokens.clj:19,44,255,319`, `reviewer.clj:361`, `tokens_test.clj:115`, `tokens_test.clj:181`, `str/upper-case` at `tokens.clj:371`.

**What I ran:** `bb` on `pr-review.tokens-test` — 22 tests, 70 assertions, 0 failures. Plus one probe script against the real namespace measuring findings 1, 2 and 4. Finding 3 is a reading of the source.

VERDICT[da5b100a]: MERGEABLE — 5 follow-ups to file

  [correctness/blocking]  none
  [coverage]              1 findings
  [style]                 none

1. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:450 — the earliest-wins rule only bounds quoted prose when the REAL banner also yields a candidate. When it yields none, the quotation is the only match and sets the park alone. MEASURED with `bb`: output whose first line is the org monthly-spend banner (which carries no `resets …` tail — `reset-at-ms` on it alone returns nil) followed by a review line quoting `resets Sep 28 at 3am (Asia/Shanghai)` gives `limited? true` and a **148.60 h** park, where the fallback would be 1 h. TRACED: `reviewer.clj:368` hands `(str (:out res) "\n" (:err res))`, and `limited?` (`tokens.clj:319-320`) is satisfied by the second pattern, which has no reset tail in it. Compounded by the `max` at `tokens.clj:488` — no later rejection can shorten it. Trigger is narrower than the pass-1 case: it needs a rejection whose banner has no parseable tail AND partial review text already on stdout. A last-match or banner-anchored parse would close both.
2. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:376 — `h12` is never checked against 1–12 before `hour` is computed. MEASURED: `resets 0am (Asia/Shanghai)` parses to midnight (1.60 h from a 22:24 now) and `resets 13am (Asia/Shanghai)` parses to 13:00 (14.60 h); only `>= 14` reaches an hour `LocalTime/of` rejects. Unreachable from the real CLI, which prints valid 12-hour values — reachable only from quoted or forged text in the reviewer's own output, where under the earliest rule it shortens a park below what the banner asked.
3. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:488 — the never-shorten `max` still makes an over-long park permanent for its full duration; there is no unpark (grep: only `park!` and `prune` write `:parked`, and `prune` drops only EXPIRED entries). Now the bound is `max-park-ms`, 8 days, not the hour this was priced at. Author has deferred it to Issue #175 (`tokens.clj:23`); listed for the ledger, not as a request to fix here.
4. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:215 — `quoted-prose-cannot-lengthen-a-park`, and its `testing` prose ("a quoted instant able only to SHORTEN a park"), state the property unconditionally, but all three assertions use `banner` = `"You have hit it · resets 3am (Asia/Shanghai)"`, which always parses. The case in finding 1 — banner contributing no candidate — is the one where the property fails, and no assertion reaches it. The test can still fail for the ordering it does check, so it is not a vacuous test, but its name certifies more than it exercises.
5. [docs-accuracy] claude-code-plugins/hooks/pr_review/tokens.clj:444 — "the worst a forged one can do is what no banner at all already does" is false in both directions, measured. Shorter: a quoted `resets 11pm` above a real `resets Sep 25 at 3am` gives a **0.60 h** park where the fallback is 1 h, so a forgery parks for less than no banner at all. Longer: finding 1's 148.60 h is 148× what no banner does. The adjacent claim at :443 ("SHORTER than the banner asked, never longer") holds only when the banner itself parses, which the sentence does not say.

yqrashawn:
## pr-review-loop — pass 3

`8fd3b7f100d1` — **MERGEABLE**

Verified by RUNNING: `bb` tokens-test (22 tests / 76 assertions, green), reviewer-test (42/137, green), `limited?` on the test file's own literal (returns true), and a reproduction of the old `hour` mapping. Everything else is a reading of the source.

Closure of the previous blocking finding — quoted prose deciding the park — confirmed at `claude-code-plugins/hooks/pr_review/tokens.clj:491-497`: the candidate is cut to `subs banner start (or nl (count banner))`, so a `resets …` on any other line contributes nothing. `tokens_test.clj:250-256` fails without it (org opening with no tail + quotation below now returns nil). Closed.

VERDICT[94724677]: MERGEABLE — 6 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  1
  [coverage]              none
  [docs-accuracy]         4
  [style]                 1

1. [correctness/followup] claude-code-plugins/hooks/pr_review/tokens.clj:492 — rule 1 assumes the org banner's reset tail sits on the same LINE as its opening clause, but `limit-patterns`:327 matches only the clause and no real org-banner sample exists anywhere in the repo (grepped: the only occurrences are invented fixtures at tokens_test.clj:174, 245, 379, and those two disagree — `·` vs `.` separator). If the real org banner puts the reset on another line or segment, `reset-at-ms` returns nil and `park!` falls back to one hour — which is exactly the measured regression this PR exists to fix (docstring:447, "an org seat parked at 22:24 whose banner said `resets 3am`"). INFERRED from the pattern's shape, not traced: no captured org banner to check against. Trigger: an org-seat monthly-spend limit whose reset is not on the clause's line. Capturing one real org banner into a fixture would settle it.
2. [docs-accuracy] claude-code-plugins/hooks/pr_review/tokens.clj:380 — "only `>= 14` was rejected downstream by `LocalTime/of`" is false. Reproduced the old `hour` cond: for `am` the branch is `hour = h12`, so 14–23 all parsed (`14am` → 14:00) and rejection began at 24; for `pm` it is `12 + h12`, so rejection began at 13. The guard at :383 is still right; the number in its justification is not.
3. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:268 — same false claim restated in the `testing` string ("only `>= 14` was caught downstream").
4. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:242 — the comment claims "this file does not contain a banner contiguously and match its own pattern". It does: `(tokens/limited? "You've hit your limit · resets 3am (Asia/Shanghai)")` returns true (ran it), and that literal is contiguous at :241, :271, :273 (plus pre-existing :369, :375). Only the org half is split. The splitting discipline the comment asserts is not actually held by this file, which is worth stating accurately since it is the property that keeps the fixture from arming the defect the deftest is about.
5. [docs-accuracy] claude-code-plugins/hooks/pr_review/reviewer.clj:362 — stale after this pass: "which `tokens/reset-at-ms` bounds by taking the EARLIEST it can see". tokens.clj:463-465 now says explicitly that EARLIEST "is NOT what closed those" and that line-anchoring is the defence against quoted text. The comment names the wrong mechanism at the one call site.
6. [style] claude-code-plugins/hooks/pr_review/tokens.clj:383 — `_ (when-not (<= 1 h12 12) (throw …))` uses a `let` binding purely for a side effect. Folding the bound into the `cond` or a `when-not` wrapper around the body expresses the same guard without the placeholder.

yqrashawn:
## Review loop summary — 3 passes

|---|---|---|---|---|---|---|
| 1 | NOT MERGEABLE | 1 | 1 | 1 | 2 | 1 |
| 2 | MERGEABLE | 0 | 3 | 1 | 1 | 0 |
| 3 | MERGEABLE | 0 | 1 | 0 | 4 | 1 |

Merging. One defect, found three times, one step further in each pass — **the parse read a different string than the gate matched**:

| pass | what set the park | measured |
|---|---|---|
| 1 | `re-find` over the whole output → a quotation *above* the banner | 138.7 h vs 18.7 h |
| 2 | earliest of all matches → when the real banner has no reset tail, the quotation is the only candidate | 138.6 h vs a 1 h fallback |
| 3 | — closed. A candidate is read only from the **line** a banner starts on | org opening + quotation below → `nil`, confirmed by the reviewer |

Two tries at "where does a candidate come from" was the tell that the answer was the wrong shape. Anchoring at the banner's *start* was also tried and is not enough: the search ran on past the banner's own line. A banner is one line.

`EARLIEST` is kept, and its justification corrected rather than restated — it is **not** what closed either measured case. It is for output carrying more than one banner line, and the docstring marks that as structural, not measured.

Also fixed across the passes: an `h12` guard (`0am` parsed as midnight, `13am` as 13:00), five stale "a park is one hour" claims swept by shape, a `12pm` branch that was never asserted while its own description claimed both ends of the clock, a test message that described its assertion backwards, and one docstring claim that was false in both directions.

Six negative controls, each armed and watched to fail. One of them **passed with the bug armed** on the first try — the two-banner fixture used a dated tail behind the subscription opening, which that pattern does not match, so there was one candidate and the rule was unexercised. The comment now says so.

## Carried to a follow-up PR

- **[correctness/followup] no real org banner exists as a fixture**, so rule 1's assumption — the org reset sits on the clause's line — is inferred from the pattern rather than traced. **I have the real sample**: this loop produced it yesterday when the org seat was actually spent, and it is one line with the reset on it. Capturing it settles the finding with evidence rather than reasoning.
- **[docs-accuracy] three false claims of my own**, all proved false by running: "only `>= 14` was rejected downstream" (it was 24 for `am` and 13 for `pm`), the same restated in a `testing` string, and a comment claiming this file never spells a banner contiguously — `limited?` returns true on its own literal, so only the org half is split. Plus a call-site comment that now names the wrong mechanism.
- **[style]** a `let` binding used purely for a side effect.

Nothing here touches [#173](https://github.com/yqrashawn/yqdotfiles/issues/173) (redaction enumerates its sources) or [#175](https://github.com/yqrashawn/yqdotfiles/issues/175) (an over-long park cannot be corrected).

## Verification

`bb test` — 301 tests, 863 assertions, 0 failures. `clj-kondo` — 0 errors.

Issue #175, opened 2026-09-22
pr-review: an over-long park cannot be corrected — there is no unpark


yqrashawn:
From #174 pass 1, `[correctness/followup]`.

`park!` refuses to shorten an existing park — two parks mean two rejections and the later instant is what both agree on, or a banner-less second rejection would replace a parsed multi-day park with an hour and hand the spent account straight back out.

The consequence is that **an over-long park cannot be corrected**. Nothing in `pr-review.tokens` unparks: only `park!` and `prune` write `:parked`, and `prune` drops expired entries only. A later, *correct* rejection is discarded by the `max`.

That was priced when a park was always one hour. It now runs to a parsed instant bounded by `max-park-ms` (8 days), so the worst case is 8 days of an account skipped.

Triggers, none of them exotic:

- the namespace docstring accepts that the clock is wall clock and that "an NTP step or a suspended host mis-times a park" — a forward step during a parsed park writes an instant that no later rejection can shorten;
- a banner whose stated reset is simply wrong, or read in a zone this machine resolves differently;
- a quoted instant that is *earlier* than the real banner shortens the park (deliberate, self-correcting) — but the reverse is now closed by taking the earliest match, so this is the remaining direction.

Degraded rather than fatal: `choose` still returns the earliest-expiring token when every token is parked, so a wrongly-parked pool never stops reviews outright — it just spends one attempt to find out.

Options worth weighing:

1. **An unpark on success.** A review that completes on a token proves the account is live; clear its park. Cheap, and it makes every over-long park self-correcting on the next successful use. The catch is that a *successful* review does not select a parked token in the first place, so this only fires where `choose` had to hand out a parked one because all were parked.
2. **Cap the never-shorten rule.** Allow a later park to shorten an existing one down to, say, `park-ms` from now, but never below. Keeps the property the `max` exists for while bounding the damage.
3. **Store the reason.** Keep the parsed instant and the banner's own text beside it, so a human reading `tokens.edn` can tell a parsed park from a fallback one — today they are indistinguishable longs.

yqrashawn:
**Won't fix** — operator's decision, 2026-09-22. Closing as accepted behaviour, not as forgotten.

What is being accepted, stated plainly so it does not have to be re-derived:

[...]

PR #176, opened 2026-09-22
fix(pr-review): capture the real org banner, and make the self-match a test

yqrashawn:
Follow-up to #174, from its pass-3 findings. *(Body rewritten after the loop — the original described an earlier shape of this branch; see the passes below.)*

## The fixture

Rule 1 of `reset-at-ms` — a reset is read only from the line its banner starts on — rested on an assumption about the **org** banner's shape that nobody had captured. `limit-patterns` matches its opening clause only, so if the real banner put its reset on another line or segment, `reset-at-ms` would return `nil` and every org-limit park would quietly be the one-hour fallback this whole change exists to replace.

The banner is now a fixture, taken from what this loop **actually received on 2026-09-21** when the account it was running on hit its limit mid-review. One line, three `·`-separated segments, reset in the last one. It parses, and `the-real-org-banner-parses` says so — with a control: move the reset to a second line and the test fails.

## The self-match, as an invariant

A file `limited?` matches is a file that, quoted back in a review, says a token is spent — and with the reset on the same line, says when it resets too.

`banner`'s comment asserted that this file never spells a banner contiguously. **It did**, and `limited?` on the file at this branch's base returns true. So the discipline is `no-plugin-file-matches-a-limit-banner` now: `limited?` over **every file a recursive glob reaches under the plugin** — 41 of them, everything but the hidden `.claude-plugin/` entries — and all the literals are assembled from halves at runtime. The control lands on `commands/review.md`, a file the first version's two-`.clj`-directory glob could not see.

It is also cwd-proof, which took a second commit. The glob was cwd-relative, and a glob that finds nothing returns an empty seq rather than throwing — so from any other cwd the whole check was vacuous and still reported 0 failures. The root is now resolved from `babashka.config` with a `user.dir` fallback and asserted to *be* the plugin; the wrong-root path reports one loud failure instead of a silent pass.

## The false claims the loop measured

- **"only `>= 14` was rejected downstream by `LocalTime/of`"** was wrong in both branches: `am` passed everything to 23 and was first rejected at **24**; `pm` was first rejected at **13**.
- **`reviewer.clj`'s park comment credited `EARLIEST`** with keeping quoted prose out. Line-anchoring does that.
- **"seven literals, one of them inside `banner` itself"** — the second half was false, and `banner`'s literal cannot match, since the pattern needs the reset tail contiguous with the clause.
- Two counts I had not measured myself, and a `"**"` inside a `testing` string that terminated it and left a bare `**` as code.

The pattern there is its own finding: every round of *rewriting* a claim produced a new wrong one, so the last round **deleted** three rather than restating them.

## Verification

`bb test` — **303 tests, 908 assertions, 0 failures**. `clj-kondo` — 0 errors. Negative controls, each armed and watched to fail: a contiguous banner in a `.md` file, the org reset moved to its own line, the 1-12 hour guard dropped, and a wrong plugin root.

The whole PR contains **one executable production change** — the 1-12 guard folded into the `cond` it guards, semantically identical to the base. Everything else is a comment or a test.

Not done, and tracked rather than hidden: `fs/glob` is eager, so a wrong root still walks the tree once before the gate (test-only, loud); `plugin-root` duplicates `hooksjson-test`'s expression; and [#173](https://github.com/yqrashawn/yqdotfiles/issues/173) / [#175](https://github.com/yqrashawn/yqdotfiles/issues/175) are untouched.


yqrashawn:
## pr-review-loop — pass 1

`88adb2db349a` — **MERGEABLE**

Reverted my armed experiment; worktree is clean again.

VERDICT[446dd19c]: MERGEABLE — 5 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  none
  [coverage]              1 findings

1. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:280 — the new invariant globs cwd-relative paths (`fs/glob "hooks/pr_review" "*.clj"`) with no non-empty guard, so from any cwd but the plugin root it contributes zero assertions and passes with the defect armed. RAN: appended one contiguous banner to `hooks/pr_review/flock.clj`; from `claude-code-plugins/` the suite reported `1 failures` with the expected message (112 assertions), from the repo root via `bb -cp claude-code-plugins/hooks:claude-code-plugins/test` the same armed tree reported `0 failures` and 78 assertions — the 34 per-file assertions (15 source + 19 test files) silently vanished because `fs/glob` on a missing root returns `0`, not a throw. The repo already has the robust pattern one namespace away: `claude-code-plugins/test/pr_review/hooksjson_test.clj:12` resolves repo files from `(System/getProperty "babashka.config")`'s parent with a `user.dir` fallback, and guards with `(is (seq hs))`. Not a verdict-flip: under the documented invocation (`docs/superpowers/plans/2026-09-08-pr-review-loop.md:133`, "`bb test` must always be run from the plugin root") the test does check the property it claims.
2. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:278 — "`limited?` on this file returned TRUE at seven literals, one of them in `banner` itself" is false in its second half. RAN: `limited?` over the base (6403e7ed7) test file returns true with exactly 7 subscription-pattern hits, at base lines 241, 271, 273, 353, 369, 372, 375 — `banner`'s own literal (base ~172) is not among them, and cannot be: the pattern needs the `resets …` tail contiguous with the clause, and there the tail is a separate argument. The PR body states both this claim and its refutation ("Re-joining `"You've hit your limit · "` inside `banner` **did not** trip it").
3. [docs-accuracy] claude-code-plugins/hooks/pr_review/reviewer.clj:363 — "a quoted whole banner LINE still can, and this exit gate is what stops it" overstates the gate. TRACED: the gate at `reviewer.clj:365` only excludes zero-exit runs; on a genuine non-zero limit rejection whose text also carries a quoted whole banner line, rule 2 (EARLIEST, `tokens.clj:499`) takes whichever reset is sooner, so a quoted line can still shorten the park — which is the direction the same diff calls "the direction that costs a review" at `tokens.clj:383`. The old comment credited EARLIEST for the wrong thing; the new one drops it entirely and leaves this residual case unnamed. Reachability is INFERRED, not traced: a `claude -p` that exits 1 on a limit prints the banner rather than a review, so the co-occurrence has not been observed.
4. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:266 — the test name ("no-source-file…") and the PR body's "runs `limited?` over every plugin source and test file" are both wider than the check: it globs `*.clj` under `hooks/pr_review` and `test/pr_review` only, not `hooks/review_core.md` (the review prompt itself), `hooks/hooks.json`, `hooks/pr-push-hook.sh`, `commands/*.md` or `skills/**`, all of which the reviewer reads and quotes just as readily. RAN: a repo-wide grep for both patterns matches nothing outside the two `.clj` files, so there is no live violation — the gap is the invariant's reach, not a current defect.
5. [style] claude-code-plugins/test/pr_review/tokens_test.clj:300 — `org`/`org-full` rebuild the org banner by hand one function below `real-org-banner`, which is now the canonical captured sample; `org-full` is `real-org-banner` with a different middle segment. Deriving the no-tail variant from the fixture would keep the two from drifting.

Verified by RUNNING: findings 1 and 2, plus the PR's own claims — `bb -e` over the `pr-review.tokens-test` namespace (24 tests, 112 assertions, 0 failures); the old hour mapping replayed standalone (`am` first rejected at 24, `14am` → 14:00, `pm` first at 13, `0am` → 00:00, matching the corrected comment at `tokens.clj:376-382`); `real-org-banner` parsing to 2026-09-22T03:00 Shanghai and to `nil` once its reset is moved to a second line (so `the-real-org-banner-parses` is a real control); `clj-kondo --lint hooks test` → `errors: 0` (33 pre-existing warnings, none from this diff). Findings 3, 4 and 5 are a reading of the source. The only executable change in the diff is folding the `1-12` guard's `_` binding into the `cond` it guards, which is semantically identical to the base — `reviewer.clj`'s hunk is comment-only.

yqrashawn:
## pr-review-loop — pass 2

`c9c158708af4` — **MERGEABLE**

Closure check against pass 1, then the sweep of `c9c158708` (test-only; `reviewer.clj`/`tokens.clj` were untouched by this commit).

```
VERDICT[446dd19c]: MERGEABLE — 5 follow-ups to file
1. [coverage] claude-code-plugins/test/pr_review/tokens_test.clj:280 — the new invariant globs cwd-relative paths …
```

- Pass-1 finding 1 (cwd-relative glob, vacuous pass) — **CLOSED**, verified by running. Armed `README.md` with one contiguous banner: from the plugin root `1 failures` (121 assertions); from the repo root via `bb --config claude-code-plugins/bb.edn` `1 failures` (121); from the repo root via `bb -cp hooks:test` the root guard fires loudly (`is not the plugin root`) instead of passing. Unarmed plugin-root run: 24 tests / 121 assertions / 0 failures. Assertion arithmetic confirms 41 real per-file checks (78 + 2 guards + 41).
- Pass-1 finding 2 (`"one of them in banner itself"`) — **CLOSED** at `tokens_test.clj:295`; the false half is gone.
- Pass-1 finding 4 (invariant narrower than its name) — **CLOSED**; now tree-wide, and the armed control lands on a non-`.clj` file the old glob could not see.
- Pass-1 finding 5 (`org`/`org-full` rebuilt by hand) — **CLOSED**; derived from `real-org-banner` by `subs`, and the `nil` control at `:350` still has no reset tail, so it remains a real control.
- Pass-1 finding 3 (`reviewer.clj:363`) — not changed; the author deferred it in the commit message with a stated reason, and the residual case is named precisely in `tokens.clj`'s own docstring. Not re-raised.

VERDICT[b8c6b6e2]: MERGEABLE — 5 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  1
  [coverage]              none
  [docs-accuracy]         3
  [style]                 1

1. [correctness/followup] claude-code-plugins/test/pr_review/tokens_test.clj:312 — both root guards are `is` forms, which do not short-circuit, so a wrong root fails ONE assertion and then still walks and slurps the whole tree anyway. RAN: `bb -cp hooks:test -e` from the repo root reports 799 assertions against 121 from the plugin root — ~719 unrelated repo files slurped after the guard had already declared the root wrong. Trigger: the namespace evaluated with no `babashka.config` from a cwd that is not the plugin root; loud, not silent, and test-only, but a `when` around the `doseq` would keep it to one line instead of an unbounded scan of whatever `user.dir` happens to be.
2. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:170 — `banner`'s docstring says `no-source-file-matches-a-limit-banner` is what holds the split-literal discipline; this same commit renamed that deftest to `no-plugin-file-matches-a-limit-banner` (`:279`). Verified by grep: `:170` is the only remaining reference and it resolves to nothing.
3. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:289 — "EVERY text file under the plugin" is not what the glob does: `fs/glob` skips hidden entries by default, so `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` are never read. RAN: `find` sees 43 files, the glob yields 41, and a probe for `.claude-plugin` in the globbed set returns `false`. The commit message's "All 41 files" is the accurate figure; the docstring's "EVERY" is not. No live violation — a repo-wide run of the same scan matched no banner anywhere.
4. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:295 — "measured on the parent commit" is now wrong for this commit's parent. RAN `limited?` over the file at all three revisions: `6403e7ed7` (the PR base) → `true`, 7 line hits; `88adb2db3` (this commit's parent) → `false`, 0 hits; HEAD → `false`. The seven were measured at the base, not at the parent. Introduced by the pass-1 fix, which replaced "measured before this test existed" with a more specific and less true phrase.
5. [style] claude-code-plugins/test/pr_review/tokens_test.clj:266 — `plugin-root` is a second copy of the expression inlined at `hooksjson_test.clj:12`, and its docstring says so ("Copied from `hooksjson-test`"). Two test namespaces now resolve the plugin root independently; a shared test helper would keep the fallback rule in one place.

Also checked and clean, so the author does not re-derive it: the PR body is stale in three places — it still names `no-source-file-matches-a-limit-banner`, still asserts the "one of them inside `banner` itself" claim that the commit message itself calls false, and reports "899 assertions" where `bb test` now gives 303 tests / 908 assertions / 0 failures. Not filed as a numbered finding because the body is not a repo path; worth a one-line edit before merge.

Verified by RUNNING: the closure of pass-1 findings 1, 2, 4 and 5, findings 1, 3 and 4 above, the full suite (303/908/0), and the armed negative control on three separate invocations. Findings 2 and 5 are a reading of the source plus a grep. `c9c158708` changes no production code — the only executable delta in the whole PR remains the `1-12` guard folded into its `cond`, which is semantically identical to the base.

yqrashawn:
## pr-review-loop — pass 3

`0eb151a41427` — **MERGEABLE**

Closure check against pass 2, then sweep of `0eb151a41` (test-only; production files untouched by this commit).

```
VERDICT[b8c6b6e2]: MERGEABLE — 5 follow-ups to file
1. [correctness/followup] claude-code-plugins/test/pr_review/tokens_test.clj:312 — both root guards are `is` forms, which do not short-circuit …
2. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:170 — `banner`'s docstring says `no-source-file-matches-a-limit-banner` …
3. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:289 — "EVERY text file under the plugin" is not what the glob does …
4. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:295 — "measured on the parent commit" is now wrong for this commit's parent …
5. [style] claude-code-plugins/test/pr_review/tokens_test.clj:266 — `plugin-root` is a second copy of the expression inlined at `hooksjson_test.clj:12` …
```

- Pass-2 #1 (`is` does not short-circuit) — **closed for the assertion explosion**, verified by running. From repo root via `bb -cp hooks:test`: `Ran 24 tests containing 80 assertions. 1 failures` with the loud `is not the plugin root` message; from plugin root 24/121/0. Residual below (finding 1): the `when` gates the `doseq`, not the walk.
- Pass-2 #2 (stale deftest name at `:170`) — **CLOSED**; `:170` now names the invariant, not a symbol. Grep: no `no-source-file-` reference left in the tree.
- Pass-2 #3 ("EVERY text file") — **CLOSED**, verified by running: glob yields exactly 41, `find` sees 43, delta is the 2 hidden `.claude-plugin/` files. Docstring at `:288` now states both.
- Pass-2 #4 ("measured on the parent commit") — **CLOSED**, verified by running `limited?` over the test file at four revisions: base `6403e7ed7` → true / 7 line hits; `88adb2db3` → false; `c9c158708` → false; HEAD → false. "on the commit this PR branched from" is the true form.
- Pass-2 #5 (`plugin-root` duplication) — not fixed; deferred in the commit message with a stated rule ("a third would be the time to extract it"). Not re-raised.
- Pass-1 #3 (`reviewer.clj:363` overstates the exit gate) — still not fixed, still deferred. Not re-raised.

Negative control re-armed at HEAD: one contiguous banner appended to `README.md` → `1 failures`, message names `README.md`. The invariant is live, not vacuous. Worktree restored.

VERDICT[fb1e29e4]: MERGEABLE — 4 follow-ups to file

  [correctness/blocking]  none
  [correctness/followup]  1
  [coverage]              none
  [docs-accuracy]         1
  [style]                 2

1. [correctness/followup] claude-code-plugins/test/pr_review/tokens_test.clj:300 — the `when` at `:324` bounds the per-file slurps but not the tree walk: `fs/glob` is eager, so `files` is fully realised at bind time, before either guard runs. RAN: `(type (fs/glob …))` → `clojure.lang.PersistentVector`, and `(count (fs/glob (System/getProperty "user.dir") "**"))` from the worktree root walked 801 entries in 5ms. Trigger unchanged from pass 2 — namespace evaluated with no `babashka.config` from a cwd that is not the plugin root; with `user.dir` at `$HOME` the walk is the home tree. Test-only, and the run already fails loudly, so this is the leftover of pass-2 #1 rather than a new defect. Binding `files` inside the same `when` (or `(when ok (->> (fs/glob …) …))`) closes it.
2. [docs-accuracy] claude-code-plugins/test/pr_review/tokens_test.clj:278 — the PR body describes a deftest that no longer exists and a scope two commits out of date: it names `no-source-file-matches-a-limit-banner` twice (renamed to `no-plugin-file-matches-a-limit-banner` at `:278` by `c9c158708`), says it "runs `limited?` over every plugin source and test file" (tree-wide, 41 files, since `c9c158708`), repeats "seven literals, one of them inside `banner` itself" two paragraphs above its own refutation, and reports "899 assertions" where `bb test` gives 908. RAN: `bb test` → `Ran 303 tests containing 908 assertions. 0 failures`. Pass 2 noted this without numbering it; filing it once so it is tracked, since three of the four claims are what a future reader would check the test against.
3. [style] claude-code-plugins/test/pr_review/tokens_test.clj:288 — the docstring reads "— 41\n of them,\n which is …": a stray two-word line left by the `**`-in-docstring fix. Also, the `41` is prose no assertion pins — the executable bound is `(< 20 (count files))` at `:316` — so the next file added under the plugin makes the docstring false with a green suite. Phrasing it as "everything the glob reaches, currently 41" would decouple it.
4. [style] claude-code-plugins/test/pr_review/tokens_test.clj:324 — `(fs/directory? (fs/path root "hooks" "pr_review"))` is written twice, at `:317` and `:324`; a `let`-bound predicate keeps the guard and the gate from diverging when one is edited.

Verified by RUNNING: the closure of pass-2 findings 1, 3 and 4; findings 1 and 2 above; the armed `README.md` control; `bb test` (303/908/0); `clj-kondo --lint hooks test` → `errors: 0, warnings: 33` (all pre-existing, none in the diff's lines). Closure of pass-2 #2 and findings 3 and 4 are a reading of the source plus grep. The whole PR still contains one executable production change — the `1-12` guard folded into its `cond` at `hooks/pr_review/tokens.clj:383`, which I read as semantically identical to the base; `reviewer.clj`'s hunk is comment-only and `0eb151a41` touches only the test file.

yqrashawn:
## Review loop summary — 3 passes (plus one MALFORMED)

|---|---|---|---|---|---|---|
| 1 | MERGEABLE | 0 | 0 | 1 | 3 | 1 |
| 2 | MERGEABLE | 0 | 1 | 0 | 3 | 1 |
| — | MALFORMED | API 500, no ledger row, retried free | | | | |
| 3 | MERGEABLE | 0 | 1 | 0 | 1 | 2 |

Merging. Closed across the passes:

- **the invariant could pass on nothing** — a cwd-relative glob that finds no files returns an empty seq, so the check was vacuous from any other cwd and still reported 0 failures. Root now resolved from `babashka.config` and asserted to *be* the plugin; the wrong-root path reports one loud failure, and the scan is bounded to 80 assertions instead of 799.
- **the invariant was narrower than its name** — two `.clj` directories became every file a recursive glob reaches (41), and the control now lands on `commands/review.md`, which the old glob could not see.
- **`org`/`org-full` rebuilt by hand** beside the captured sample — derived from it now, so they cannot drift from the one fixture whose whole point is that the shape is not invented.
- **three false claims of mine**, each measured false by the reviewer.

The instructive part: every round of *rewriting* a claim produced a new wrong one — a renamed deftest referenced by its old name, "EVERY text file" where the glob skips hidden entries, "the parent commit" after a commit moved the parent. The last round **deleted** three claims instead of restating them. One more of the same kind was caught by the suite rather than a reviewer: a `"**"` inside a `testing` string terminated it and left a bare `**` as code, in a docstring about globbing.

## Not done, tracked not hidden

- **[correctness/followup] `fs/glob` is eager**, so a wrong root still walks the tree once before the gate closes. Test-only, and the run already fails loudly. Binding `files` inside the `when` closes it; not worth a fourth pass on a test file.
- **[style]** `plugin-root` duplicates `hooksjson-test`'s expression, and the `41` in the docstring is prose no assertion pins. Two copies; a third is when to extract.
- **[docs-accuracy, pass 1 #3]** `reviewer.clj`'s comment does not name the case where a quoted whole banner *line* arrives inside a genuine rejection and shortens the park — [#175](https://github.com/yqrashawn/yqdotfiles/issues/175)'s territory, and the reviewer marked its reachability inferred.

[#173](https://github.com/yqrashawn/yqdotfiles/issues/173) untouched.

## Verification

`bb test` — 303 tests, 908 assertions, 0 failures. `clj-kondo` — 0 errors. One executable production change in the whole PR: the 1-12 hour guard folded into the `cond` it guards.

